PR #6, opened 2018-05-12
Move repetitive content to detours.h under #ifdef DETOURS_INTERNAL.

jaykrell:


dtarditi:
This change looks fine, other than the DETOURS_INTERNAL #ifdef being at the top of Detours.h.     It could confuse people reading the header file to see this as the first thing.  Could you move the code lower in Detours.h, where other DETOURS_INTERNAL code is located.

jaykrell:
Just move into the existing slightly later DETOURS_INTERNAL block? Yes that looks like it will work.

dtarditi:
Yes, just moving it slightly later would be great.

jaykrell:
If you move it down just one "section" you get:

z:\s\detours\src\detours.h(47) : fatal error C1189: #error :  Unknown architecture (x86, amd64, ia64, arm, arm64)
modules.cpp

because _X86_ etc. are defined by windows.h.

I probably put this right at the top, because everyone else was doing it before including detours.h.

jaykrell:
Prefer repetition and prettier detours.h? Or just delayed?

dtarditi:
Could we leave the #include "windows.h" in the .c files, and the move the code down in Detours.h?   Or does  the #define _ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE need to precede the include of "windows.h"?

jaykrell:
I think what you hint at is true -- _ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE must be early. At least in the past. ARM development was by default limited to "the store", unless explicitly enable "desktop" development.
Oh, um, another option is /D_ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE in Makefile. I like source to be more build-system-independent though.

PR #9, opened 2018-05-12
ARM64 does not like addresses below 4GB -- rebase tests higher.

jaykrell:
Part of these changes:

Change 114680 by NTDEV\jaykrell@JAYKRELL100-4 on 2017/10/20 22:24:17
        ARM64 does not like addresses below 4GB.
        Disable managed test on ARM64 until/unless anyone constructs an adequate environment.
        It is disabled on ARM32 also.
Affected files ...
... //depot/969/private/jaykrell/3.0/samples/einst/Makefile#5 edit
... //depot/969/private/jaykrell/3.0/samples/findfunc/Makefile#4 edit
... //depot/969/private/jaykrell/3.0/samples/Makefile#4 edit

Change 114788 by NTDEV\jaykrell@JAYKRELL100-4 on 2017/11/16 01:00:07
        fix copy/pasto that breaks all except arm64
Affected files ...
... //depot/969/private/jaykrell/3.0/samples/findfunc/Makefile#5 edit

### DIFF
--- a/samples/einst/Makefile
+++ b/samples/einst/Makefile
@@ -9,6 +9,14 @@
 
 !include ..\common.mak
 
+# ARM64 does not like base addresses below 4GB.
+
+!if "$(DETOURS_TARGET_PROCESSOR)" == "ARM64"
+DETOURS_BASE_SUFFIX=00
+!else
+DETOURS_BASE_SUFFIX=
+!endif
+
 LIBS=$(LIBS) kernel32.lib user32.lib
 
 all: dirs \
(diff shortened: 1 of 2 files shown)


jaykrell:
It is just doing math. One address is the other shifted left 8.

dtarditi:
I understand what it is doing.  I'd like you to add makefile variables that hold the base addresses and set them conditionally based on the target architecture.  I think that'll be clearer in the long run than adding suffixes to link addresses. 

I also disabled compiling these tests in the recent baseline that I committed.  Could you reenable compiling these tests and enable them for tests, if they work.

Thanks.

PR #10, opened 2018-05-12
Expand ARM64 rbCode (and scratch) to 128 and document why.

jaykrell:
This is a little high, but 64 seemed low.

Change 114686 by NTDEV\jaykrell@JAYKRELL100-4 on 2017/10/23 19:06:28
        Expand ARM64 rbCode (and scratch) to 128 and document why.
        This is a little high, but 64 seemed low.
Affected files ...
... //depot/969/private/jaykrell/3.0/src/detours.cpp#27 edit
... //depot/969/private/jaykrell/3.0/src/disasm.cpp#68 edit

### DIFF
--- a/src/disasm.cpp
+++ b/src/disasm.cpp
@@ -313,7 +313,7 @@ class CDetourDis
 
     LONG                m_lScratchExtra;
     PBYTE               m_pbScratchTarget;
-    BYTE                m_rbScratchDst[64];
+    BYTE                m_rbScratchDst[64]; // matches or exceeds rbCode
 };
 
 PVOID WINAPI DetourCopyInstruction(_In_opt_ PVOID pDst,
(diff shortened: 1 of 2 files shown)


src/disasm.cpp
```diff
@@ -313,7 +313,7 @@ class CDetourDis
 
     LONG                m_lScratchExtra;
     PBYTE               m_pbScratchTarget;
-    BYTE                m_rbScratchDst[64];
+    BYTE                m_rbScratchDst[64]; // matches or exceeds rbCode
```
dtarditi: rbCode is 128 bytes, and this is only 64 bytes.

src/disasm.cpp
```diff
@@ -2547,7 +2547,7 @@ class CDetourDis
     PBYTE   m_pbPool;
     LONG    m_lExtra;
 
-    BYTE    m_rbScratchDst[64];
+    BYTE    m_rbScratchDst[64]; // matches or exceeds rbCode
```
dtarditi: rbCode is 128 bytes and this is only 64 bytes.

jaykrell:
Each architecture is different. Double check?

PR #17, opened 2018-05-13
Remove opcode fields from COPYENTRY as it isn't used, except

jaykrell:
for a self-check.

### DIFF
--- a/src/disasm.cpp
+++ b/src/disasm.cpp
@@ -198,60 +198,60 @@ class CDetourDis
   protected:
 // These macros define common uses of nFixedSize, nFixedSize16, nModOffset, nRelOffset, nFlagBits, pfCopy.
 #define ENTRY_DataIgnored           0, 0, 0, 0, 0,
-#define ENTRY_CopyBytes1            1, 1, 0, 0, 0, &CDetourDis::CopyBytes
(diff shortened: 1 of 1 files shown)


src/disasm.cpp
```diff
@@ -886,542 +885,542 @@ const BYTE CDetourDis::s_rbModRm[256] = {
 
 const CDetourDis::COPYENTRY CDetourDis::s_rceCopyTable[257] =
```
jaykrell: How about we remove this explicit 257, remove the terminal entry, and C_ASSERT that the size is 256? That gives some of the sanity check, w/o requiring the tedious recording of the numbers, aside from comments which have less stringent requirement.

dtarditi:
This seems good.  There are some merge conflicts that need to be resolved first.

dtarditi:
@jaykrell, could you resolve the merge conflicts?

dtarditi:
@jaykrell, gentle reminder: could you resolve the merge conflicts?

jaykrell:
I redid the change fairly mechanically, with regexp search/replace, mostly, rather than resolve the conflict.
This should retain confidence.

bgianfo:
@JayKrell can you rebase this on master again, so we can get CI build validation?

bgianfo:
@dtarditi You ok with merging this now?

PR #24, opened 2018-07-12
Fix 4GB offset checks in UPDATE_IMPORTS_XX/FindAndAllocateNearBase

austinkinross:
On x64, imports are found via a 32-bit offset from the start of the image. As a result, when Detours replaces imports on x64, it must make sure that the offsets to the new imports fit into a 32-bit variable.

Before this change, FindAndAllocateNearBase tried to validate this requirement, but its logic wasn't quite right. It used its pbBase argument, which is strictly larger than the start of the image, to validate that the offset from pbBase to the its new allocation fitted into 32-bits. This could fail if:
   1) pbBase wasn't page-aligned AND
   2) pbBase fell in a pure free region

This is because FindAndAllocateNearBase could decide to allocate at the page boundary below pbBase, since VirtualQueryEx rounds its inputted address down to its next page boundary. This meant that pbAddress - pbBase could be a very large number, which would fail the old 4GB checks in FindAndAllocateNearBase.

The fix for this issue is to move the 4GB checks up into FindAndAllocateNearBase's caller, i.e. UPDATE_IMPORTS_XX. This allows us to check the exact values that must fit into 32-bits, instead of incorrectly checking offsets from pbBase.

This change ultimately allows us to inject our DLLs into some video games with unusual size values in their _IMAGE_OPTIONAL_HEADER structs. pbBase is calculated using these sizes, so these unusual sizes resulted in pbBase's value satisfying both of the conditions mentioned above. This resulted in us failing the old 4GB checks in FindAndAllocateNearBase, which meant that we failed to inject our DLLs into the process.

### DIFF
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -180,16 +180,6 @@ static PBYTE FindAndAllocateNearBase(HANDLE hProcess, PBYTE pbBase, DWORD cbAllo
 
         PBYTE pbAddress = (PBYTE)(((DWORD_PTR)mbi.BaseAddress + 0xffff) & ~(DWORD_PTR)0xffff);
 
-#ifdef _WIN64
-        // The distance from pbBase to pbAddress must fit in 32 bits.
-        //
-        const size_t GB4 = ((((size_t)1) << 32) - 1);
-        if ((size_t)(pbAddress - pbBase) > GB4) {
-            DETOUR_TRACE(("FindAndAllocateNearBase(1) failing due to distance >4GB %p\n", pbAddress));
-            return NULL;
-        }
-#endif
-
         DETOUR_TRACE(("Free region %p..%p\n",
                       mbi.BaseAddress,
                       (PBYTE)mbi.BaseAddress + mbi.RegionSize));
@@ -201,14 +191,7 @@ static PBYTE FindAndAllocateNearBase(HANDLE hProcess, PBYTE pbBase, DWORD cbAllo
                 DETOUR_TRACE(("VirtualAllocEx(%p) failed: %d\n", pbAddress, GetLastError()));
                 continue;
             }
-#ifdef _WIN64
-            // The distance from pbBase to pbAddress must fit in 32 bits.
-            //
-            if ((size_t)(pbAddress - pbBase) > GB4) {
-                DETOUR_TRACE(("FindAndAllocateNearBase(2) failing due to distance >4GB %p\n", pbAddress));
-                return NULL;
-            }
-#endif
+
             DETOUR_TRACE(("[%p..%p] Allocated for import table.\n",
                           pbAlloc, pbAlloc + cbAlloc));
             return pbAlloc;
--- a/src/uimports.cpp
+++ b/src/uimports.cpp
@@ -147,6 +147,20 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
     PIMAGE_IMPORT_DESCRIPTOR piid = (PIMAGE_IMPORT_DESCRIPTOR)pbNew;
     DWORD_XX *pt;
 
+#ifdef _WIN64
+    // The offset from pbModule to any replacement import must fit into 32 bits.
+    // For simplicity, we check that the offset to the last byte fits into 32 bits,
+    // instead of the largest offset we'll actually use. The values are very similar.
+    const size_t GB4 = ((((size_t)1) << 32) - 1);
+    const size_t maxPossibleOffset = (size_t)(pbNewIid + cbNew - 1 - pbModule);
+
+    if (maxPossibleOffset > GB4) {
+        DETOUR_TRACE(("Offset into the allocation for new imports is >4GB %zx\n",
+                      maxPossibleOffset));
+        goto finish;
+    }
+#endif
+
     DWORD obBase = (DWORD)(pbNewIid - pbModule);
     DWORD dwProtect = 0;
 


src/creatwth.cpp
```diff
@@ -180,16 +180,6 @@ static PBYTE FindAndAllocateNearBase(HANDLE hProcess, PBYTE pbBase, DWORD cbAllo
 
         PBYTE pbAddress = (PBYTE)(((DWORD_PTR)mbi.BaseAddress + 0xffff) & ~(DWORD_PTR)0xffff);
```
dtarditi: I'm not convinced that the problem is that VirtualQueryEx is rounding pbBase down to the nearest page boundary.  I think there may be another issue going on here.

If the problem is VirtualQueryEx rounding down to the nearest page boundary, then line 181 should fix this. It rounds pbAddress up to the nearest 64K boundary.  Assuming that pages are less than 64K in size, pbAddress would be at or above pbBase.   (I don't recall whether Windows x64 supports large virtual page sizes (2Mbyte) that are supported by the x64 architecture, in which case this reasoning is flawed).

I suspect the problem is that mbi.BaseAddress may be rounded down even further.  The only guarantee from VirtualQueryEx is that lpAddress (the address for which we're querying information about) is within the memory range starting at mbi.BaseAddress.   Suppose  that mbi.BaseAddress is aligned to a memory manager allocation granularity and <  pbBase.  Then line 181 doesn't do anything, and the existing check fails.

Of course, mbi.RegionSize could be even lower, depending on what the memory manager decided to do.   In this case, pbAlloc could be < pbBase.

src/uimports.cpp
```diff
@@ -147,6 +147,20 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
     PIMAGE_IMPORT_DESCRIPTOR piid = (PIMAGE_IMPORT_DESCRIPTOR)pbNew;
     DWORD_XX *pt;
 
+#ifdef _WIN64
+    // The offset from pbModule to any replacement import must fit into 32 bits.
+    // For simplicity, we check that the offset to the last byte fits into 32 bits,
+    // instead of the largest offset we'll actually use. The values are very similar.
+    const size_t GB4 = ((((size_t)1) << 32) - 1);
+    const size_t maxPossibleOffset = (size_t)(pbNewIid + cbNew - 1 - pbModule);
```
dtarditi: This is assuming that pbNewIiid >= pbModule.

jaykrell: It is supposed to be. I'd have to read more surrounding code to be sure but it is certainly supposed to be.

src/creatwth.cpp
jaykrell: Independing of all the rounding concerns, the original check, that I put in, is overly strict in the sense that Austin is experiencing.

The allocation and certain bytes within it must be within 4GB of the start of the image.
The check however checks for roughly 4GB from the end of the image -- it doesn't compute the end of the image very well imho, but it tries. It uses end of the image, because that is where it tries to start allocating from.

This was my mistake, a few years ago.

Historically the check wasn't even there.

And on desktop it rarely mattered -- for the CreateProcessWithDll case -- it'd always succeed.
I added the check, because on some Windows variants, sections (images) and virtualmemory occupy separate and far apart ranges (more than 4GB apart), and the address "successfully" allocated here was never within 4GB. And it'd get silently truncated, and then crash in ntdll.dll using the truncated offset.

src/uimports.cpp
dtarditi: I agree that the check needs to be updated.   We certainly need to have a check on x64 to avoid random crashs, as you point out.

src/creatwth.cpp
austinkinross: Hi David, 

My understanding of VirtualQueryEx's behavior is based on [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/aa366907(v=vs.85).aspx), which says that "[lpAddress] is rounded down to the next page boundary". This description matches the behavior I saw while debugging the issue, but I wouldn't know if it can act differently in other cases.

> If the problem is VirtualQueryEx rounding down to the nearest page boundary, then line 181 should fix this. It rounds pbAddress up to the nearest 64K boundary. Assuming that pages are less than 64K in size, pbAddress would be at or above pbBase.

I believe you're correct, unless pbBase lies in the first page after a 64K boundary. In that case, VirtualQueryEx would round it down to the 64K boundary, and line 181 would assign pbAddress to the same 64K boundary. pbAddress would then be less than pbBase, making us fail the old 4GB checks. 

Indeed, this is exactly what I see when repro'ing this issue. I've copied the relevant code below, annoted with extra bolded comments describing misc variable values when the issue occurs. 

It sounds like I may have missed a third precondition for pbBase in my PR comment: that is, pbBase needs to lie in the first page after a 64K boundary.

Thanks for your help.

<pre>
static PBYTE FindAndAllocateNearBase(HANDLE hProcess, PBYTE pbBase, DWORD cbAlloc)
{
    <b>//
    // pbBase = 0x00007ff79a920800
    //</b>
 
    MEMORY_BASIC_INFORMATION mbi;
    ZeroMemory(&mbi, sizeof(mbi));
 
    PBYTE pbLast = pbBase;
    for (;; pbLast = (PBYTE)mbi.BaseAddress + mbi.RegionSize) {
 
        ZeroMemory(&mbi, sizeof(mbi));
        if (VirtualQueryEx(hProcess, (PVOID)pbLast, &mbi, sizeof(mbi)) == 0) {
            if (GetLastError() == ERROR_INVALID_PARAMETER) {
                break;
            }
            DETOUR_TRACE(("VirtualQueryEx(%p) failed: %d\n",
                          pbLast, GetLastError()));
            break;
        }
 
        <b>
        //
        // According to MSDN, VirtualQueryEx rounds pbLast down to its nearest page boundary.
        // As a result, mbi.BaseAddress = 0x00007ff79a920000.
        // </b>

        // Usermode address space has such an unaligned region size always at the
        // end and only at the end.
        //
        if ((mbi.RegionSize & 0xfff) == 0xfff) {
            break;
        }
 
        // Skip anything other than a pure free region.
        //
        if (mbi.State != MEM_FREE) {
            continue;
        }
       
        <b>
        //
        // Since 0x00007ff79a920000 is 64KB-aligned, the following line just sets pbAddress = 0x00007ff79a920000
        //</b>
        PBYTE pbAddress = (PBYTE)(((DWORD_PTR)mbi.BaseAddress + 0xffff) & ~(DWORD_PTR)0xffff);
 
#ifdef _WIN64
        // The distance from pbBase to pbAddress must fit in 32 bits.
        //
        const size_t GB4 = ((((size_t)1) << 32) - 1);
        if ((size_t)(pbAddress - pbBase) > GB4) {
            <b>
            //
            // The above statement is true. pbAddress = 0x00007ff79a920000 and pbBase = 0x00007ff79a920800,
            // so “pbAddress – pbBase” is a very large number when casted to size_t.
            // FindAndAllocateNearBase returns NULL as a result, and DetourUpdateProcessWithDll fails.
            //</b>
            DETOUR_TRACE(("FindAndAllocateNearBase(1) failing due to distance >4GB %p\n", pbAddress));
            return NULL;
        }
#endif
                …
}
</pre>

austinkinross:
Thank you David for the feedback, and thanks for the clear instructions. I've updated the pull request.

PR #26, opened 2018-07-14
Added vc sln and vcxproj files

OfekShilon:
Currently required PlatformSDK is default (8.1), considering requiring the one bundled with latest VS (10.0.17134.0).

Included configurations for Release/Debug x MT/MD x x64/x86 .

This can save some mundane work for devs working in VS, wishing to use detours. It's time to modernize the build script a bit anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,11 @@
+﻿################################################################################
+# This .gitignore file was automatically created by Microsoft(R) Visual Studio.
+################################################################################
+
+*.ipch
+*.suo
+/.vs
+*.user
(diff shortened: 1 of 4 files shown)


LonghronShen:
Maybe PR #48 also helps.
Using CMake to generate VS projects looks good, or just open the CMakeLists.txt using VS2017+ is better.

dtarditi:
@OfekShilon, thanks for the proposed PR.  What do you think of the suggestion from @LonghronShen to use CMake instead?

OfekShilon:
Sounds like a good idea.

PR #27, opened 2018-07-17
Add DetourAllocateRegionWithinJumpBounds API

msmania:
This patch exports the ability to allocate a region within the jump bounds from the given address because this ability is useful outside Detours.  The logic was already implemented as a part of the internal function
`detour_alloc_trampoline`.  The change is to extract it out as a public function `DetourAllocateRegionWithinJumpBounds`.

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -1250,6 +1250,65 @@ static PVOID detour_alloc_region_from_hi(PBYTE pbLo, PBYTE pbHi)
     return NULL;
 }
 
+static PVOID detour_alloc_trampoline_allocate_new(PBYTE pbTarget,
+                                                  PDETOUR_TRAMPOLINE pLo,
+                                                  PDETOUR_TRAMPOLINE pHi)
+{
+    PVOID pbTry = NULL;
+
+    // NB: We must always also start the search at an offset from pbTarget
+    //     in order to maintain ASLR entropy.
+
+#if defined(DETOURS_64BIT)
+    // Try looking 1GB below or lower.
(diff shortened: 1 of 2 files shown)


dtarditi:
Generally, we're trying to avoid expanding the functionality of Detours.  We want it to stay focused on the core problem that it solves.   This change seems reasonable, though, because it is simply opening an API as public.

I noticed when reviewing the code that ` DetourAllocateRegionWithinJumpBounds` tries to allocate a region that is a multiple of system page granularity (4 Kbytes).   This led to changes in the functions that it calls.   On Windows, the Virtual Memory subsystem won't allocate virtual memory regions at a granularity of less than 64 KBytes of memory anway.  Is there a reason why you are trying to allocate memory at a smaller granularity?

msmania:
I removed the second commit bringing 4K granularity part and rebased the branch.  Now the change is to export the existing logic as an API without any behavior change.

sonyps5201314:
> On Windows, the Virtual Memory subsystem won't allocate virtual memory regions at a granularity of less than 64 KBytes of memory anway.

I think that's wrong,In my actual tests, I found that the minimum allocation was 4KB.
```cpp
	PBYTE setup = (PBYTE)detour_alloc_region_from_lo(pbLo, pbHi, 17, TRUE);
	ATLASSERT(setup);
	while (setup< pbLo)
	{
		setup = (PBYTE)detour_alloc_region_from_lo(setup+ 0x10000, pbHi, 17, TRUE);
		MessageBox(NULL,_T("TEST"),_T("APP"),MB_OK);
		if (setup == NULL)
		{
			// Out of memory.
			return NULL;
		}
	}
```
image

PR #30, opened 2018-08-01
Establish ARM64 baseline.

dtarditi:
This establishes a baseline for ARM64 of compiling/non-compiling tests and working/non-working tests for ARM64.  There are a number of pull requests with bug fixes and this will let us validate them.  Presumably we'll be able to move more tests to working status after the bug fixes.

This also add a missing function detour_jmp_find_bounds for ARM64.  Jay, I just copied the 32-bit ARM one.  If there's a better fix, let me know.

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -974,6 +974,19 @@ inline PBYTE detour_skip_jmp(PBYTE pbCode, PVOID *ppGlobals)
     return pbCode;
 }
 
+inline void detour_find_jmp_bounds(PBYTE pbCode,
+                                   PDETOUR_TRAMPOLINE *ppLower,
+                                   PDETOUR_TRAMPOLINE *ppUpper)
+{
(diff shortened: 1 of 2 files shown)


src/detours.cpp
```diff
@@ -974,6 +974,19 @@ inline PBYTE detour_skip_jmp(PBYTE pbCode, PVOID *ppGlobals)
     return pbCode;
 }
 
+inline void detour_find_jmp_bounds(PBYTE pbCode,
```
jaykrell: I wonder if we shoudn't have common code:
`detour_find_jmp_within_2gb` or `detour_find_jmp_within_2gb_above_or_below` or `detour_find_jmp_boudnds_common`?

jaykrell: I think Pedro's PR adds what you have here?

jaykrell: Typo in commit message (sorry).

PR #36, opened 2018-08-27
Update README

unknownv2:
* Update information for licenses and backwards compability


unknownv2:
Sorry about that... meant to merge to fork instead of original. Delete if possible!

PR #41, opened 2018-11-05
Fix C4819 warning in non-english build environment

mrsshr:
The '±' character can not be correctly represented in some code pages.
This causes the build to fail with a C4919 warning.

This commit replaces the '±' character in the comments with '+/-'.


dtarditi:
Looks good, thanks!

PR #48, opened 2018-12-10
[WIP] Feature: Adding CMake support

LonghronShen:
Try to add a simple CMakeLists.txt for building detours in x86/x64.
Support for other architectures like ARM/ARM64/IA64  may be added later.


dtarditi:
Thanks for the proposed change.  I have some suggestions.

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+SET(LIB_PATCH_VERSION "1")
+SET(LIB_VERSION_STRING "${LIB_MAJOR_VERSION}.${LIB_MINOR_VERSION}.${LIB_PATCH_VERSION}")
+
+# compile in release with debug info mode by default
+IF(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel." FORCE)
+ENDIF()
+
+FILE(GLOB detours_src
+    ${CMAKE_CURRENT_LIST_DIR}/src/*.h
```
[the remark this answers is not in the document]
LonghronShen: Thank you for your suggestions!
Since I'm quite familiar with the Detours itself, I just use the script then it builded successfully. I will listing all these files here.
Then you're right, it is not nesscessary to add the header files. I'll remove the pattern.

LonghronShen:
Since the `uimports.cpp` file is included in `creatwth.cpp#L347`, I suggest that the `uimports.cpp` file should be renamed to `uimports.h`, then we can use `*.cpp` pattern to add all source files in the CMakeLists.txt.

schlamar:
It looks like Detours relies on build order of sources to compile reliable (AFAIS detours.cpp should be first). However, this is pretty bad as VS doesn't guarantee build order of source files.

I assume this is the reason your CMake file doesn't work for me. I guess your CMake configuration works for just one specific build order which I'm not getting on my machine. You are defining a lot of stuff which shouldn't be defined at all (e.g. DETOURS_VERSION, _X86_, DETOURS_X86, etc.). 

Please note that the nmake Makefile is only passing `/DWIN32_LEAN_AND_MEAN /D_WIN32_WINNT=0x501` to the compiler. (DETOURS_BITS=X is passed to rc.exe, but this is only relevant on building the examples).

My own CMake configuration is working with these two definitions, too. I have put detours.cpp at first entry in the source list. However, I'm not sure if this is reproducible or just luck.

I guess the correct approach would be to refactor Detours so it isn't dependent on build order anymore.

LonghronShen:
I can't agree more! About the cross platform/architecture problem, we also need a more uniform design.

dtarditi:
@longhronshen, what is the status of this change?  We do need a cmake configuration to work for different developers.  Are there dependencies missing from the cmake file that would better express the build order?

LonghronShen:
Now the PR has been done for the main project, and several other example projects.
What is still blocked is some example projects which involves some COM interops.
If this is not the key problem, I can update the PR and then I think we can merge the PR.

dtarditi:
I think it is fine to exclude some samples for now from the cmake build.  We can merge the PR and that can be worked on later.  Please make sure it is documented that those samples don't have a cmake build yet.

LonghronShen:
# CMake support

CMake support for the main project has been done for a while, but some sample projects are excluded for some interop reasons.

* cping
* disas
* dtest
* dumpe
* dumpi
* echo
* einst
* excep
* findfunc
* impmunge
* Makefile
* member
* opengl
* slept
* talloc
* tryman

Other TODOs:

* ARM/ARM64/IA64 support
* Improve Sample project debugging experiences

dtarditi:
Please let me know when this is ready for further review.

LonghronShen:
I think the current version is ready for review, except for these excluded sample projects.

CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+ELSE(WIN32)
+    MESSAGE(FATAL_ERROR "Only Win32 platforms are supported.")
+ENDIF(WIN32)
+
+# TODO: What about Windows on ARM, or IA64 systems?
+IF("${CMAKE_SIZEOF_VOID_P}" EQUAL "8")
+	LIST(APPEND DETOURS_COMPILE_DEFINITIONS "DETOURS_TARGET_PROCESSOR=X64")
+	LIST(APPEND DETOURS_COMPILE_DEFINITIONS DETOURS_X64)
+	LIST(APPEND DETOURS_COMPILE_DEFINITIONS DETOURS_64BIT)
+	LIST(APPEND DETOURS_COMPILE_DEFINITIONS _AMD64_)
```
bgianfo, line 61: The MSVC compiler sets these based on the architecture the configured compiler is targeting, the build system shouldn't be setting it them selves.

cmake/FindCSharp.cmake
```diff
@@ -0,0 +1,87 @@
+##
+## $Id$
+## It is part of the SolidOpt Copyright Policy (see Copyright.txt)
+## For further details see the nearest License.txt
+##
+
+## This file is based on the work of SimpleITK:
+##  https://github.com/SimpleITK/SimpleITK/blob/master/CMake/FindCSharp.cmake
```
bgianfo, line 8: We can't copy and paste code from other projects into Detours, this code has it's own LICENSE.

CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
+CMAKE_MINIMUM_REQUIRED(VERSION 3.10)
```
bgianfo, line 1: Why is everything in all caps? can we fix it to be lower case?

bgianfo:
Can you fix the merge conflicts in .gitignore?

LonghronShen:
OK, I'll take it in the next one or two days.

CMakeLists.txt
LonghronShen: As for older cmake code style, commands are written in UPPER CASE.

cmake/FindCSharp.cmake
LonghronShen: For this, I'll see if we can solve the problem by referencing the third party staffs as submodule or some other ways...

CMakeLists.txt
LonghronShen: This is just imitating the actions written in Makefiles. Or can we change the code so that we can use modern ways to check these staffs?

bgianfo:
@LonghronShen can you enable "Allow edits from Maintainer" for this PR?
See: https://github.blog/2016-09-07-improving-collaboration-with-forks/

I have some fixes / additions locally that I would like to include before we merge this PR.

Thanks!

CMakeLists.txt
bgianfo: I can run https://github.com/cheshirekow/cmake_format on the files if I can get push access to this PR.

cmake/FindCSharp.cmake
bgianfo: CMake 3.8+ has native support for C#
See: https://cmake.org/cmake/help/latest/release/3.8.html#id4

jpr89:
Any update on this?

tostercx:
Any chance of this getting merged?

bgianfo:
Yea it's on my backlog of things to do.

Since the author doesn't seem available, I'm planning on cherry-picking their commits and then fixing them up, and pushing it once it's ready. Let me know if you are able / willing to help with that.

sylveon:
Preferably we should use a single build system for the entire project, because maintaining several is a maintenance burden more than anything else. Since CMake can generate Visual Studio solutions, maybe ditch nmake and the .sln?

bgianfo:
> Preferably we should use a single build system for the entire project, because maintaining several is a maintenance burden more than anything else. Since CMake can generate Visual Studio solutions, maybe ditch nmake and the .sln?

Yup, completely agree, and that's why I haven't merged this yet.
I want it to be feature complete before we merge and use it the one true build system. 
CMake can generate NMake as well, so there will be no reason to keep the nmake or sln after this is merged.

defaultactor:
last post was Sep 13, 2021, is this still an ongoing effort or has cmake been abandoned?

jpr42:
I've made https://github.com/microsoft/Detours/pull/390 based on the feedback in this PR and my knowledge of modern CMake. I also did my best to add fair bit of testing.

We use detours in several projects for open source and at my company so we would like to get proper CMake support

jpr42:
> last post was Sep 13, 2021, is this still an ongoing effort or has cmake been abandoned?

Looks like this effort was abandoned so I made a new one.

> > Preferably we should use a single build system for the entire project, because maintaining several is a maintenance burden more than anything else. Since CMake can generate Visual Studio solutions, maybe ditch nmake and the .sln?
> 
> Yup, completely agree, and that's why I haven't merged this yet. I want it to be feature complete before we merge and use it the one true build system. CMake can generate NMake as well, so there will be no reason to keep the nmake or sln after this is merged.

I've done my best to make the new CMake branch feature complete @bgianfo

jpr42:
I ran a fair bit of testing on my fork to make sure everything was working properly

jpr42:
As well as locally of course

PR #55, opened 2019-03-03
Add a new sample dynamic_alloc to test the new API

msmania:
This patch adds a new sample to test a new API `DetourAllocateRegionWithinJumpBounds` that was introduced by #27.  I'll update the wiki after this PR is done.

### DIFF
--- a/samples/dynamic_alloc/main.cpp
+++ b/samples/dynamic_alloc/main.cpp
@@ -1,0 +1,177 @@
+//////////////////////////////////////////////////////////////////////////////
+//
+//  Detours Test Program
+//
+//  Microsoft Research Detours Package
+//
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//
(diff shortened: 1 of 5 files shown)


samples/dynamic_alloc/main.cpp
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  template <typename T>
+  static T *at(void *base, uint32_t offset) {
+    return
+      reinterpret_cast<T*>(
+        reinterpret_cast<uint8_t*>(base) + offset);
+  }
+
+  void *region_{};
```
dtarditi: Please avoid the C++ {} style initializers in the sample code.  There are a few other places where you've used that style.

msmania: Thank you for the comments.  I added more comments and replaced initialization code.

msmania:
Thanks. I've updated wiki.

PR #57, opened 2019-03-21
Replace ARM64's patch jmp from a 16 byte encoding to a 12 byte encoding, as some functions in Windows are only 12 bytes long and have no padding.

pmsjt:


src/detours.cpp
```diff
@@ -929,6 +930,79 @@ inline void write_opcode(PBYTE &pbCode, ULONG Opcode)
...
+        ULONG V : 1;
+        ULONG iop2 : 3;
+        ULONG size : 2;
+    } ldr;
+
+    ULONG br;
+};
+
+#pragma warning(push)
+#pragma warning(disable:4201)
```
jaykrell: nit: I like comments on these but ok. `// nonstandard extension`.

src/detours.cpp
```diff
@@ -929,6 +930,79 @@ inline void write_opcode(PBYTE &pbCode, ULONG Opcode)
...
+    // ldr x17, [x17, jmpval]
+    // br x17
+
+    struct ARM64_INDIRECT_JMP *pIndJmp;
+    union ARM64_INDIRECT_IMM jmpIndAddr;
+
+    jmpIndAddr.value = (((LONG64)pbJmpVal) & 0xFFFFFFFFFFFFF000) - 
+                       (((LONG64)pbCode) & 0xFFFFFFFFFFFFF000);
+
+    pIndJmp = (struct ARM64_INDIRECT_JMP *)pbCode;
```
jaykrell: nit: Remove `struct`?

src/detours.cpp
```diff
@@ -1062,7 +1136,10 @@ inline void detour_find_jmp_bounds(PBYTE pbCode,
                                    PDETOUR_TRAMPOLINE *ppLower,
                                    PDETOUR_TRAMPOLINE *ppUpper)
 {
-    // We have to place trampolines within +/- 2GB of code.
+    // The encoding used by detour_gen_jmp_indirect actually enables a
+    // displacement of +/- 4GiB. In the future, this could be changed to
```
jaykrell: nit: I'd just say `4GB`, not sure what people are disambiguating with the larger term.

src/detours.cpp
```diff
@@ -929,6 +930,79 @@ inline void write_opcode(PBYTE &pbCode, ULONG Opcode)
...
+#pragma warning(pop)
+
+PBYTE detour_gen_jmp_indirect(BYTE *pbCode, ULONG64 *pbJmpVal)
+{
+    // adrp x17, [jmpval]
+    // ldr x17, [x17, jmpval]
+    // br x17
+
+    struct ARM64_INDIRECT_JMP *pIndJmp;
+    union ARM64_INDIRECT_IMM jmpIndAddr;
```
jaykrell: Nits: remove `struct` and `union`. The `typedef` is implied and heavily dependend upon in C++.

src/detours.cpp
```diff
@@ -910,12 +910,13 @@ struct _DETOUR_TRAMPOLINE
     _DETOUR_ALIGN   rAlign[8];      // instruction alignment array.
     PBYTE           pbRemain;       // first instruction after moved code. [free list]
     PBYTE           pbDetour;       // first instruction of detour function.
+    ULONG64         rbIndTarget;
 };
 
-C_ASSERT(sizeof(_DETOUR_TRAMPOLINE) == 184);
+C_ASSERT(sizeof(_DETOUR_TRAMPOLINE) == 192);
```
jaykrell: I think, part of the point, but not the main point, is this can now be reduced.

jaykrell: Oh, maybe not. The large increase is not the jmp per se, but the other copied instructions?

src/detours.cpp
```diff
@@ -1683,7 +1760,8 @@ LONG WINAPI DetourTransactionCommitEx(_Out_opt_ PVOID **pppFailedPointer)
 #endif // DETOURS_ARM
 
 #ifdef DETOURS_ARM64
-            PBYTE pbCode = detour_gen_jmp_immediate(o->pbTarget, NULL, o->pTrampoline->pbDetour);
+            o->pTrampoline->rbIndTarget = (ULONG64)(o->pTrampoline->pbDetour);
```
jaykrell: nit: no parens needed on `o->pTrampoline->pbDetour`.

jaykrell: Same on next line. I get it, though, `->` is an operator, with such high precedence people don't even see it as such.

src/detours.cpp
pmsjt: Interestingly enough, even CLANG and GCC seem to support this. But, yes, these are nonstandard extensions.

src/detours.cpp
```diff
@@ -929,6 +930,79 @@ inline void write_opcode(PBYTE &pbCode, ULONG Opcode)
     pbCode += 4;
 }
 
+struct ARM64_INDIRECT_JMP {
+    struct {
```
jaykrell: This is all cryptic to me but that is probably inevitable.

pmsjt:
> 
> 
> Nitty feedback, mainly to remove excess `struct` and `union`.
> I don't really know if this works, but surely you got it right.
> Were you able to run the unit tests?
> I'd have to regain a Windows/arm64 machine. I used Nebula before.

Yup - there are ISVs (such as Samsung SDS) running production code on this change since last year.

Thanks Jay.

src/detours.cpp
```diff
@@ -929,6 +930,79 @@ inline void write_opcode(PBYTE &pbCode, ULONG Opcode)
...
+    };
+
+    LONG64 value;
+};
+
+#pragma warning(pop)
+
+PBYTE detour_gen_jmp_indirect(BYTE *pbCode, ULONG64 *pbJmpVal)
+{
+    // adrp x17, [jmpval]
```
dtarditi: This doesn't quite match the ARM assembler syntax.  It should be // adrp x17, jmpval,  where jmpval is interpreted as a label, i.e. an immediate.    I thinhk it'd be helpful to add a comment to the effect calculating the address of jmpval is split across two instructions.  The first one adrp (loads most significant bits) the second ldr (load the least significant bits of jmpval).

pmsjt: Sure, we can add a comment here about the split loading. However this is the canonical and standard way of loading quantities bigger than a page. That's just how things are always done on ARM64. So is it worth it?

jaykrell: This is just, "fixed size 32bit instructions require multiple instructions to load a 32bit or 64bit immediate"? Yeah, seems obvious. Maybe some day we will forget x86 and doing it in a single instruction will seem weird. :)

src/detours.cpp
```diff
@@ -929,6 +930,79 @@ inline void write_opcode(PBYTE &pbCode, ULONG Opcode)
...
+
+    pIndJmp->ardp.Rd = 17;
+    pIndJmp->ardp.immhi = jmpIndAddr.adrp_immhi;
+    pIndJmp->ardp.iop = 0x10;
+    pIndJmp->ardp.immlo = jmpIndAddr.adrp_immlo;
+    pIndJmp->ardp.op = 1;
+
+    pIndJmp->ldr.Rt = 17;
+    pIndJmp->ldr.Rn = 17;
+    pIndJmp->ldr.imm = (((ULONG64)pbJmpVal) & 0xFFF) / 8;
```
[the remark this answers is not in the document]
pmsjt: This is the address of rbIndTarget which is a ULONG64. This has to be an 8-byte aligned field as we are loading an 8-byte quantity.

src/detours.cpp
```diff
@@ -910,12 +910,13 @@ struct _DETOUR_TRAMPOLINE
     _DETOUR_ALIGN   rAlign[8];      // instruction alignment array.
     PBYTE           pbRemain;       // first instruction after moved code. [free list]
     PBYTE           pbDetour;       // first instruction of detour function.
+    ULONG64         rbIndTarget;
```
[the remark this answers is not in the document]
pmsjt: Will do.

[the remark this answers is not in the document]
pmsjt: This field was really redundant with pbDetour, as both were holding the same value.

src/detours.cpp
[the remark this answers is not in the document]
dtarditi: I'm concerned that someone will pass in an improperly aligned value for pbJmpVal.  That's why I was suggest an assert.

dtarditi:
@pmsjt, I had one suggestion for adding an assert.  This is your call about whether to add it.

PR #59, opened 2019-05-03
Update README.TXT

waawaa:
This is a common error when trying to compile detours from CLI


sylveon:
I think a VS build environment is assumed by this, since you can't even use nmake without using the developer command prompt.

dtarditi:
I've opened PR #61 to improve the README file, which subsumes this PR.

dtarditi:
PR #61 took care of the issue covered by this PR.  Thanks for the suggestion.

PR #75, opened 2019-08-17
Fixes for a few subtle problems

zeffy:
Here is a breakdown of this PR by commit

### Fix incorrect usage of ZeroMemory
`ZeroMemory` was being given the size of the pointer, rather than the size of the struct.

### Fix inconsistent declaration of DetourFindFunction
`DetourFindFunction` was declared as using `LPCSTR`, but the actual definition of the function uses `PCSTR`. Not a huge problem, but it causes a warning in some compilers.

### Move DETOURS_STRINGIFY macros, and fix inconsistent reference to them
There was couple few different issues here:

- `VER_DETOUR_BITS` macro was referencing a non-existent macro `DETOUR_STRINGIFY`
- `DETOURS_STRINGIFY` was using `DETOURS_STRINGIFY_` before it was declared

I fixed both by reordering the macro declarations and having `VER_DETOUR_BITS` use `DETOURS_STRINGIFY`.

### Fix macros in detver.h
Same as previous but for the macros in `detver.h`


src/detours.h
```diff
@@ -543,8 +543,8 @@ PVOID WINAPI DetourSetSystemRegionUpperBound(_In_ PVOID pSystemRegionUpperBound)
 
 ////////////////////////////////////////////////////////////// Code Functions.
 //
-PVOID WINAPI DetourFindFunction(_In_ LPCSTR pszModule,
-                                _In_ LPCSTR pszFunction);
+PVOID WINAPI DetourFindFunction(_In_ PCSTR pszModule,
```
dtarditi: Could you change the definition of the function to match the declaration instead?  You altered the declaration.   LPCSTR has slightly better information for static analysis.    From documentation:
```
LPCSTR

This type is declared in WinNT.h as follows:
typedef __nullterminated CONST CHAR *LPCSTR;
```
versus:
```
PCSTR

This type is declared in WinNT.h as follows:
typedef CONST CHAR *PCSTR;
```
LPCSTR has a __nullterminated SAL annotation on it.

zeffy: That's a good point, I corrected it in the latest commits.

PR #80, opened 2019-08-26
Fix: Drop non-pseudo handles to current thread in DetourUpdateThread

### LINKED ISSUE #78: DetourUpdateThread with non-pseudo current thread handle will freeze the thread.
When running the following code, detours will suspend the current thread, with no way of resuming it, possibly resulting in a process deadlock if this is done in DllMain.

```cpp
HANDLE hThread = OpenThread(THREAD_SUSPEND_RESUME, false, GetCurrentThreadId());
DetoursUpdateThread(hThread);
// no code will execute after this point.
```

This is caused by https://github.com/microsoft/Detours/blob/master/src/detours.cpp#L1854 only checking against the pseudo-handle instead of also checking if the TID of the handle matches the current thread's.
This if check can be safely replaced by `GetThreadId(hThread) == GetCurrentThreadId()` to handle both real and pseudo handles to the current thread (although there might be a slight performance gain due to less API calls by still explicitly checking if it's a pseudo handle)

sylveon:
Fixes #78


poizan42:
I think NtQueryInformationThread with ThreadBasicInformation should give you the thread id in THREAD_BASIC_INFORMATION.ClientId.UniqueThread on Windows XP.

sylveon:
Yeah that was a solution, but `THREAD_BASIC_INFORMATION` is undocumented (it probably doesn't matter on XP since that is not getting any changes but I'd like to know if it's OK to use undocumented functions)

bgianfo:
Another possible way to fix this is to duplicate the thread handle, and check the duplicated handle.
So we know we never are checking a pesudo handle.

sylveon:
The problem is specifically in regards to things that aren't a pseudo handle though. Currently this is blocked on the fact that Detours needs to target Windows XP (GetThreadId was added in Vista)

sonyps5201314:
after official merge my PR[#127](https://github.com/microsoft/Detours/pull/127), DetourUpdateAllOtherThreads will instead of the old DetourUpdateThread API, and it will exclude currrent thread automaticall. and DetourUpdateAllOtherThreads is a safer solutuon for HOOK operation. and CreateToolhelp32Snapshot mentioned above is slowly. It is not fit for extreme environment.

sylveon:
`CreateToolhelp32Snapshot` is fine, I was doing this in a `DllMain` so any new thread would have been blocked from running by the loader lock.

bgianfo:
> We should at least document this behavior in the wiki, saying that passing a non-pseudo handle to the current thread is 
> unsupported and will result in freezing.
> 
> While at it, also maybe document that passing a pseudo handle is a noop? I've seen many examples do that even if it's useless

@sylveon Good suggestion, I've added documentation for both of these.
https://github.com/microsoft/Detours/wiki/DetourUpdateThread

sonyps5201314:
> `CreateToolhelp32Snapshot` is fine, I was doing this in a `DllMain` so any new thread would have been blocked from running by the loader lock.
Not use is a speed problem, not a function problem. This is just for efficiency reasons.
you can understand why replace CreateToolhelp32Snapshot with ZwQuerySystemInformation in the follow patch of Mhook
https://github.com/apriorit/mhook/commit/f1b41f3d939ec2e424c40bf35cff4b4dce9152b9

bgianfo:
I propose we go forward with this change, but just have it just fall back to the previous behavior on systems without the `GetThreadId()` export. I want to get this into the next release, if folks running on XP need this fix they can submit an appropriate fix for XP. It probably makes sense to factor this out into a function so this logic, and potentially future XP specific logic can be contained there.

Thoughts?

sylveon:
I can get this working this weekend

sylveon:
I updated the code to dynamically try loading it now. Also as a drive-by I changed `IsWow64ProcessHelper` to use `GetModuleHandleW` because `LoadLibraryW` "leaks" (even though practically it's not an issue since kernel32.dll is always loaded)

sylveon:
Although, I guess #141 will also resolve that - is the intent still to merge that PR for next release? I can revert that specific change to avoid conflicts with it.

justanotheranonymoususer:
> Not use `CreateToolhelp32Snapshot` is a speed problem, not a function problem. This is just for efficiency reasons.
> you can understand why replace CreateToolhelp32Snapshot with ZwQuerySystemInformation in the follow patch of Mhook
> [apriorit/mhook@f1b41f3](https://github.com/apriorit/mhook/commit/f1b41f3d939ec2e424c40bf35cff4b4dce9152b9)

@sonyps5201314 you got me interested, so I checked it. See my comment here:
https://github.com/apriorit/mhook/commit/f1b41f3d939ec2e424c40bf35cff4b4dce9152b9#commitcomment-56063743

In short, I found that both methods are almost equally slow. That's unfortunate, I wish there was a quick way to get the thread list of only a single process. The kernel can get this information, see e.g. the implementation of `NtSuspendProcess`, but as far as I know it doesn't expose it.

PR #81, opened 2019-08-26
Feature: Add DetourFindRemotePayload and improve other payload-related methods

### LINKED ISSUE #79: Feature: DetourFindRemotePayload and DetourFindRemotePayloadEx for remote process.
In my use-case, it's possible that a DLL gets injected and uninjected multiple times in the target process's lifetime. I want to add a payload to the target process for my DLL to check if it is in the target process rather than the injecting process which uses SetWindowsHookEx to ensure that the DLL is uninjected by the system if my process dies because the detour is used to transmit information to my application, so it is not useful (or cannot live at all) if my application is not running. SetWindowsHookEx requires that I need to load the DLL in the injecting process.

Since it's possible for the DLL to be uninjected, calling DetourCreateRemotePayload each time it needs to be injected will pollute the process with multiple copies of the magic payload. I cannot use a static boolean variable to track if the remote payload exists, because the case where the DLL gets injected or uninjected is either my process or the target process starting or closing (my process restarting would reset that boolean variable)

Currently, I would need to copy DetourFindPayload's implementation to my own code as well as using CreateToolhelp32Snapshot (or use ReadProcessMemory and read the module headers) to know if the payload already exists in the target process.

This should be a feature of Detours and the signature of the two functions should be the same than the in-process ones, except the void pointer returned is suitable for further reading or writing via ReadProcessMemory or WriteProcessMemory. This would have the side effect of allowing some sort of live configuration system for the detour, where the detour can just read configuration from the payload and the injector can write to it to dynamically change the behavior of the detour instead of having to establish another method of communication with the injector process (which would have been more complex for both parties).

sylveon:
Other improvements are:
- Makes the `pcbData` parameter in DetourFindPayload and DetourFindPayloadEx optional, so that if an application only needs to search for the presence of a payload, they can ignore the size by passing `nullptr`.
- Makes the `pvData` parameter in DetourCopyPayloadToProcess const, so that a pointer to a const C++ object can be passed instead of the object needing to be `const_cast`ed or being non-const.
- Adds DetourCopyPayloadToProcessEx, which has the same interface than DetourCopyPayloadToProcess, but it returns the address of the payload in the remote module, if the program later wants to write to it.

Fixes #79

From basic testing, it seems to work for me. I can apply corrections and add a sample if required (although finding a trivial use case for those APIs would be required)


sylveon:
@dtarditi could this and other PRs/issues be addressed? I'd like at some point to start using this in production (it's been working flawlessly in development)

dtarditi:
@sylveon, what is the motivation for wanting to add DetourFindRemotePayload?   The other changes look straightforward.   It isn't clear to me how this new function would help with the core goal of detours, which is allow ppeople to instrument and monitor and API calls.

sylveon:
My use-case is described in #79.

Basically, I use the payload as a flag to check if the DLL is actually loaded in the process I want to detour API calls (much like the current CreateProcessWithDll code does) and I may use it in the future to transmit information to the DLL when it is injected (like a shared handle or a pointer to a memory map).

However, the way my app works, the user can close and reopen it at will, so the DLL and detour may be removed (as in, completely unloaded from its address space) and added multiple times within the lifetime of the target process.

Without this API, I cannot know if a payload already exists in the target process, so I always have to create a new payload, making its potential future purpose of transmitting information useless since there is not guarantee about which payload DetourFindPayload will give you if there are multiple payloads with the same GUID, and it also leaks memory each time my app is closed and relaunched due to Detours not providing a way to deallocate the payload from the target process.

With this API, I can instead just find an existing payload (or create a new payload if there is none yet), not having to allocate a new one each time, and then edit the payload's content using WriteProcessMemory calls.

This function can very well be written standalone, but since it taps into implementation details of Detours, which might change in the future, I think that having it as part of the core codebase would be better.

sylveon:
Note that my commits became unverified due to me changing my GPG key, you can ignore that.

dtarditi:
@sylveon, this use case makes sense.  Could you add a test case for the new API function?

sylveon:
@dtarditi I fixed a bug related to memory allocation (used `delete` instead of `delete[]`, fixed that issue by avoiding heap allocations entirely), and added a sample. The sample is not representative of my real world use, but it should be good enough to test and demonstrate the API. What it does is:
- creates a child process
- creates a payload in that child process containing a handle of the parent process
- parent process adds a payload to itself containing zeroes using `DetourCopyPayloadToProcessEx` and remembers the memory address of the payload data
- child process grabs the handle to the parent process using `DetourFindPayloadEx`
- child process finds the payload in the parent process using `DetourFindRemotePayload` and fills it with random data using `WriteProcessMemory`
- child process exits using the same random data as exit code
- parent process makes sure the payload data matches the child process exit code, and returns with 1 if it doesn't

src/creatwth.cpp
```diff
@@ -118,13 +123,148 @@ static HMODULE WINAPI EnumerateModulesInProcess(HANDLE hProcess,
...
 }
 
+//////////////////////////////////////////////////////////////////////////////
+//
+// Find payloads in target process.
+//
+
+static PVOID FindDetourSectionInRemoteModule(HANDLE hProcess,
+                                             HMODULE hModule,
+                                             PIMAGE_NT_HEADERS32 pNtHeader,
```
bgianfo, line 146: Can you add SAL annotations to the new functions please? 
https://docs.microsoft.com/en-us/cpp/code-quality/annotating-function-parameters-and-return-values?view=vs-2019

src/modules.cpp
```diff
@@ -824,7 +824,7 @@ _Writable_bytes_(*pcbData)
 _Readable_bytes_(*pcbData)
 _Success_(return != NULL)
 PVOID WINAPI DetourFindPayloadEx(_In_ REFGUID rguid,
-                                 _Out_ DWORD * pcbData)
+                                 _Out_opt_ DWORD *pcbData)
```
bgianfo, line 827: Do you want to send these as separate PR's so we can merge them faster?

bgianfo:
It you don't mind, it would be nice if you could include the markdown for the WIKI documentation pages for these new API's in this PR (just in the PR description, or as comments) we can just paste them in once we merge.

samples/payload/Makefile
```diff
@@ -0,0 +1,59 @@
...
+##  Makefile for Detours Test Programs.
+##
+##  Microsoft Research Detours Package
+##
+##  Copyright (c) Microsoft Corporation.  All rights reserved.
+##
+
+!include ..\common.mak
+
+CFLAGS=$(CFLAGS) /std:c++17 /EHsc
```
bgianfo, line 12: People routinely report and fix issues compiling detours with MSVC 2005 and other crazy old configurations, how do I'm worried about taking a dependency on modenish c++, I agree it would be nice but not sure how realistic it is.

samples/payload/payload.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+int main()
+{
+	std::wstring target = GetProcessFileName(GetCurrentProcess());
+	target.erase(target.rfind(L'\\') + 1);
+	target += L"payloadtarget.exe";
+
+	STARTUPINFOW si = { sizeof(si) };
+	PROCESS_INFORMATION pi;
+	winrt::check_bool(CreateProcessW(target.c_str(), nullptr, nullptr, nullptr, false, CREATE_SUSPENDED, nullptr, nullptr, &si, &pi));
```
bgianfo, line 38: winrt::check_bool throws, seems like we are missing a try/catch block to report failure?

sylveon: If an uncaught exception is thrown, the process will automatically exit with an non-0 code. Since cppwinrt needs to go away (because it requires C++17) I'll look into adding some logging.

samples/payload/Makefile
sylveon: I added that for cppwinrt but I'll remove it.

sylveon: Should work with old compilers now

bgianfo:
@dtarditi, do you think this needs more work? Or is it ready to merge?

src/modules.cpp
```diff
@@ -789,23 +789,12 @@ PVOID WINAPI DetourFindPayload(_In_opt_ HMODULE hModule,
...
-                pSection->guid.Data4[6] == rguid.Data4[6] &&
-                pSection->guid.Data4[7] == rguid.Data4[7]) {
-
+            if (DetourAreSameGuid(pSection->guid, rguid)) {
                 if (pcbData) {
                     *pcbData = pSection->cbBytes - sizeof(*pSection);
-                    SetLastError(NO_ERROR);
-                    return (PBYTE)(pSection + 1);
                 }
+                SetLastError(NO_ERROR);
```
bgianfo, line 796: I think this is going to get fixed in #139 FYI, and will lead to merge conflict.

src/detours.cpp
```diff
@@ -2545,4 +2545,20 @@ BOOL WINAPI DetourVirtualProtectSameExecute(_In_  PVOID pAddress,
                                              pAddress, nSize, dwNewProtect, pdwOldProtect);
 }
 
+BOOL WINAPI DetourAreSameGuid(_In_ REFGUID left, _In_ REFGUID right)
+{
+    return
```
bgianfo, line 2550: I realize  this is just a nice refactor of the existing code, but is there a reason not use IsEqualGUID that I didn't notice?

sylveon: Using `IsEqualGUID` makes detours depend on ole32.lib.

bgianfo: Uhh.. maybe they changed it in different versions of the headers...

In my windows SDK (version 10.0.19041.0) `IsEqualGUID` is declared as a macro or inline function depending on how you are compiling. 

```cpp

#if !defined (__midl)
#if !defined (_SYS_GUID_OPERATORS_)
#define _SYS_GUID_OPERATORS_
#include <string.h>

// Faster (but makes code fatter) inline version...use sparingly
#ifdef __cplusplus
__inline int InlineIsEqualGUID(REFGUID rguid1, REFGUID rguid2)
{
   return (
      ((unsigned long *) &rguid1)[0] == ((unsigned long *) &rguid2)[0] &&
      ((unsigned long *) &rguid1)[1] == ((unsigned long *) &rguid2)[1] &&
      ((unsigned long *) &rguid1)[2] == ((unsigned long *) &rguid2)[2] &&
      ((unsigned long *) &rguid1)[3] == ((unsigned long *) &rguid2)[3]);
}

__inline int IsEqualGUID(REFGUID rguid1, REFGUID rguid2)
{
    return !memcmp(&rguid1, &rguid2, sizeof(GUID));
}

#else   // ! __cplusplus

#define InlineIsEqualGUID(rguid1, rguid2)  \
        (((unsigned long *) rguid1)[0] == ((unsigned long *) rguid2)[0] &&   \
        ((unsigned long *) rguid1)[1] == ((unsigned long *) rguid2)[1] &&    \
        ((unsigned long *) rguid1)[2] == ((unsigned long *) rguid2)[2] &&    \
        ((unsigned long *) rguid1)[3] == ((unsigned long *) rguid2)[3])

#define IsEqualGUID(rguid1, rguid2) (!memcmp(rguid1, rguid2, sizeof(GUID)))

#endif  // __cplusplus
```

sylveon: Ah, I was going off the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/guiddef/nf-guiddef-isequalguid)... We just need to know since when that was made inline. For example, I'm not sure if whatever SDK supported by VS 2005 also does that :/

bgianfo: Sure, it looks like it still exports IsEqualGUID... 
```
$ dumpbin /exports C:\Windows\System32\ole32.dll | rg IsEqualGUID
       1139  140 00010CA0 IsEqualGUID
```

I think your change is correct, maybe we should just add a comment to
DetourAreSameGuid, explaining that it exists for backwards compat.

src/modules.cpp
sylveon: I'll locally merge once the other PR is merged

src/detours.cpp
sylveon: I added a comment

src/modules.cpp
sylveon: Seems like this didn't even cause a merge conflict, since the code diff was the exact same.

sylveon:
Any update on this?

bgianfo:
@dtarditi waiting on your feedback here.

dtarditi (approved):
Looks good to me.

bgianfo:
@sylveon I would like to include this in the next release can you: 
- [ ] Fix up the merge conflicts
- [ ] Add some basic unit test coverage of this in the new [unit tests](https://github.com/microsoft/Detours/tree/master/tests).

sylveon:
Yeah I was actually planning to take a look at the conflicts tonight. I'll see for the unit tests as well.

sylveon:
I resolved conflicts and added unit test @bgianfo 
While I was at it I noticed quite a few random tabs where present in source code, so I find & replace'd them, which will help in preparation to #181

bgianfo:
Sorry can you submit the tab changes as a seperate PR? Since the repro is configured to only allow squash commits, that's going to make the commit diff pretty messy. ☹

sylveon:
Reverted that commit for now

sylveon:
There seems to be some contamination between PR builds, the 2019 builds are complaining about tests for DetourFindPayloadEx already existing in test_modules_api.cpp at line 623, but the file isn't even that long in my branch.

sylveon:
Looks like I've got conflicts to resolve anyways - will do tonight

bgianfo (approved):

bgianfo:
Thanks again for the contribution @sylveon!

bgianfo:
I merged all of the doc changes that you provided in your gist as well. 
See: https://github.com/microsoft/Detours/wiki/_compare/6c0fef6d375e2eebf558be1f6c3c8a817a4121a4...da14a2afdadfa3c7a4018b7e44485fb2825ce00c

PR #96, opened 2020-01-11
Fix dynamic_alloc to build if UNICODE is defined.

jaykrell:
People often copy Detours to other build environments, that
might define UNICODE by default.


sylveon:
The proper fix for this would be to use the TEXT or _T macro in the call
sites of Log.

jaykrell:
Not really. Only if you are actually going to compile twice. Pick just one way preferably.

PR #99, opened 2020-01-20
Replace implicit type conversion with incorrect semantics by proper return value

michael-os:
The dynamic_alloc example fails to compile on Visual Studio 2015 due to an C4800 / C2220 compiler error (LONG status is used as return value). Apart from the performance warning, which is treated as error because of the /WX flag, the semantics of this operation seem to be wrong. At the end of the function, we find `return status == NO_ERROR`. That means that in line 70 a successful execution results in a return value of `true` and an unsuccessful execution results in `false`. However, if the function is left in line 52, the return code is `true` (because the test `status != NO_ERROR` was positive).

On the other hand, the return value of the method is nowhere used, so the function signature might just as well be adjusted to use void at this moment.

### DIFF
--- a/samples/dynamic_alloc/main.cpp
+++ b/samples/dynamic_alloc/main.cpp
@@ -49,7 +49,7 @@ bool DetourTransaction(std::function<bool()> callback) {
   LONG status = DetourTransactionBegin();
   if (status != NO_ERROR) {
     Log("DetourTransactionBegin failed with %08x\n", status);
-    return status;
+    return status == NO_ERROR;
   }
 
   if (callback()) {
--- a/samples/common.mak
+++ b/samples/common.mak
@@ -23,7 +23,7 @@ CLIB=/MT
 !ENDIF
 
 AFLAGS=/nologo /Zi /c /Fl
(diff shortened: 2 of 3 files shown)


samples/dynamic_alloc/main.cpp
```diff
@@ -49,7 +49,7 @@ bool DetourTransaction(std::function<bool()> callback) {
   LONG status = DetourTransactionBegin();
   if (status != NO_ERROR) {
     Log("DetourTransactionBegin failed with %08x\n", status);
-    return status;
+    return status == NO_ERROR;
```
bgianfo, line 52: I think we should just return `false` explicitly, the current suggested change can potentially raise a different static analysis warning, since the comparison is guaranteed to always be false.

bgianfo:
Thanks for catching this error, and for submitting the patch @michael-os!
If you don't mind addressing the minor comments I left, I'd be happy to merge.

samples/dynamic_alloc/main.cpp
bgianfo: I can fix later, I'll just merge as is... :)

bgianfo (approved):
I can fix later, thanks!

PR #102, opened 2020-03-31
Add vcpkg installation instructions

NancyLi1013:
`Detours `is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `detours `and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `detours`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (Windows: x86, x64, ARM, UWP) to keep a wide coverage for users.
Note: Currently only support dynamic build on Windows.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/detours/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/samples/README.TXT
+++ b/samples/README.TXT
@@ -33,6 +33,18 @@ samples, type "nmake".
 Note that you must build setdll and syslog in order to use many of the
 other sample programs.
 
+INSTALLING AND BUILDING VIA VCPKG:
+==================================
+You can download and install detours using the vcpkg(https://github.com/Microsoft/vcpkg) dependency manager:
+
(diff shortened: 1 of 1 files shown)


sylveon:
I wrote that script!

As for dynamic builds, Detours as a library only supports static linking.

NancyLi1013:
Thanks for your correct info about dynamic build.
In fact, Detours only supports static linkage in vcpkg no matter what is built as dynamic or static.

NancyLi1013:
Thanks for your help.

PR #108, opened 2020-05-19
Allow compilation with MinGW to succeed

virtuald:
What the title says. MinGW doesn't support `__try` and `__except`, so it's just defined out here as it doesn't seem there's a better way to handle it.

Note: Includes #107

### DIFF
--- a/src/detours.h
+++ b/src/detours.h
@@ -48,6 +48,12 @@
 #pragma warning(pop)
 #endif
 
+#ifdef __GNUC__
+#define __try
+#define __except(x) if (0)
+#include <strsafe.h>
+#endif
+
 #endif // DETOURS_INTERNAL
 
 //////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


bgianfo:
Without SEH support, it's not safe to execute the code that is currently guarded by __try / __except.
You'll crash at runtime, you should be able to easily verify this by running any of the tests, they will most likely crash instantly.

I'm not sure it makes sense to even support detours being built and run in this configuration.

bgianfo:
Also, it looks like the changes from #107 where mistakenly included here as well.

Thaodan:
How did you compile it with mingw, by hand or using different makefiles?

virtuald:
Using different makefiles. I used cmake, very simple.

bgianfo:
Merging this as it appears harmless and enables this corner case scenario. 
Thanks @virtuald sorry for the delay!

PR #111, opened 2020-08-02
Update createwth.cpp with enum values.

lironzua:
Replace CLR flags magic numbers with enum values instead.

### DIFF
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -478,7 +478,7 @@ static BOOL UpdateFrom32To64(HANDLE hProcess, HMODULE hModule, WORD machine,
     }
 
     // Remove the import table.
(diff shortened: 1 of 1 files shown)


bgianfo:
Thanks for the patch @lironzua !

Would you mind rebasing this on top of the latest master, so we can get build validation before merge?

bgianfo (changes requested):
Please rebase

bgianfo (approved):

bgianfo:
I re-based it my self.

The only thing I worry about here is someone is the potential that someone is going to be building on a very old Windows SDK that could potentially not include these headers. If that happens to be the case, then I suppose we can re-define the constants ourselves.

PR #112, opened 2020-08-02
Replace DWORDXX with IMAGE_THUNK_DATAXX.

lironzua:
Since both of the types have the same capacity in the same arch, they both work. But IMAGE_THUNK_DATAXX is the right struct for this code and I believe to make the code more readable and descriptive.

### DIFF
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -323,6 +323,7 @@ static BOOL RecordExeRestore(HANDLE hProcess, HMODULE hModule, DETOUR_EXE_RESTOR
 #define IMAGE_NT_HEADERS_XX             IMAGE_NT_HEADERS32
 #define IMAGE_NT_OPTIONAL_HDR_MAGIC_XX  IMAGE_NT_OPTIONAL_HDR32_MAGIC
 #define IMAGE_ORDINAL_FLAG_XX           IMAGE_ORDINAL_FLAG32
+#define IMAGE_THUNK_DATAXX              IMAGE_THUNK_DATA32
 #define UPDATE_IMPORTS_XX               UpdateImports32
 #define DETOURS_BITS_XX                 32
 #include "uimports.cpp"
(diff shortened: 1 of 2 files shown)


bgianfo (changes requested):
Hey @lironzua would you mind rebasing this on latest master so we can make sure it doesn't break the new build validation that was added before merge?

lironzua:
@bgianfo done

lironzua:
My bad, I updated something else by mistake. I rebased again here. Also, what PR validation did you mean? LMK if anything is missing. 

Thanks

bgianfo:
>  Also, what PR validation did you mean? LMK if anything is missing.

Looks like it caught a build failure. 
See: https://github.com/microsoft/Detours/pull/112/checks?check_run_id=1074164464

bgianfo:
Did you mean to use `IMAGE_THUNK_DATA` instead of `IMAGE_THUNK_DATAXX`? I can't find any evidence of `IMAGE_THUNK_DATAXX` existing in my local copy of the Windows 10 SDK for instance.

src/creatwth.cpp
```diff
@@ -323,6 +323,7 @@ static BOOL RecordExeRestore(HANDLE hProcess, HMODULE hModule, DETOUR_EXE_RESTOR
 #define IMAGE_NT_HEADERS_XX             IMAGE_NT_HEADERS32
 #define IMAGE_NT_OPTIONAL_HDR_MAGIC_XX  IMAGE_NT_OPTIONAL_HDR32_MAGIC
 #define IMAGE_ORDINAL_FLAG_XX           IMAGE_ORDINAL_FLAG32
+#define IMAGE_THUNK_DATAXX           	IMAGE_THUNK_DATA32
```
bgianfo, line 326: nit:  can you fix the indention to match the surrounding?

bgianfo (approved):

PR #113, opened 2020-08-02
Add documentation

lironzua:
Add an explanation on why we need to deal with a missing IAT.

### DIFF
--- a/src/uimports.cpp
+++ b/src/uimports.cpp
@@ -91,10 +91,12 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
 
         DETOUR_TRACE(("ish[%d] : va=%08x sr=%d\n", i, ish.VirtualAddress, ish.SizeOfRawData));
         
-        // If the linker didn't suggest and IAT in the data directories, the loader will look for the
-        // section of the import directory to be used for this instead. Since we put out new IMPORT_DIRECTORY
-        // outside any section boundary, the loader will not find it. So we give one explicitly to avoid
-        // the search.
+        // If the linker didn't suggest and IAT in the data directories, the
+        // loader will look for the section of the import directory to be used
+        // for this instead. Since we put out new IMPORT_DIRECTORY outside any
+        // section boundary, the loader will not find it. So we provide one
+        // explicitly to avoid the search.
+        //
         if (inh.IAT_DIRECTORY.VirtualAddress == 0 &&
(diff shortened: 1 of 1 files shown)


src/uimports.cpp
```diff
@@ -90,8 +90,11 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
         }
 
         DETOUR_TRACE(("ish[%d] : va=%08x sr=%d\n", i, ish.VirtualAddress, ish.SizeOfRawData));
-
-        // If the file didn't have an IAT_DIRECTORY, we assign it...
+        
+        // If the linker didn't suggest and IAT in the data directories, the loader will look for the
+        // section of the import directory to be used for this instead. Since we put out new IMPORT_DIRECTORY
+        // outside any section boundary, the loader will not find it. So we give one explicitly to avoid
```
bgianfo, line 96: > So we give one explicitly to avoid

Small nitpick: I think "provide" would be better than "give".

src/uimports.cpp
```diff
@@ -90,8 +90,11 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
         }
 
         DETOUR_TRACE(("ish[%d] : va=%08x sr=%d\n", i, ish.VirtualAddress, ish.SizeOfRawData));
-
-        // If the file didn't have an IAT_DIRECTORY, we assign it...
+        
+        // If the linker didn't suggest and IAT in the data directories, the loader will look for the
+        // section of the import directory to be used for this instead. Since we put out new IMPORT_DIRECTORY
```
bgianfo, line 95: nit: 
Would you mind reformating the comments so they obey the 80 character line limit convention? 
It seems to be obeyed in majority of the detours code base, we should try to keep things consistent.

bgianfo:
Thanks for the patch @lironzua! 
If you could just address the minor nits, I'll be happy to merge.

bgianfo (changes requested):
Thanks for the patch @lironzua! 
If you could just address the minor nits, I'll be happy to merge.

src/uimports.cpp
bgianfo: I pushed a fix for this.

src/uimports.cpp
bgianfo: I pushed a fix for this

bgianfo (approved):

bgianfo:
I fixed the minor nit picks, and pushed them, thanks again @lironzua!

PR #117, opened 2020-08-07
Add test case for mov reg, imm64 but it already worked, bug fixed many years ago.

jaykrell:


samples/disas/disas.cpp
```diff
@@ -407,6 +407,25 @@ int WINAPI WinMain(HINSTANCE hinst, HINSTANCE hprev, LPSTR lpszCmdLine, int nCmd
...
+    //
+    {
+        static const UCHAR mov_r10_imm64[] = {0x49, 0xba, 1, 2, 3, 4, 5, 6, 7, 8 };
+
+        PVOID const after = DetourCopyInstructionX64(0, 0, const_cast<PUCHAR>(mov_r10_imm64), 0, 0);
+
+        if (after != &mov_r10_imm64 + 1)
+        {
+            printf("mov_r10_imm64 failed, expected:%p vs. got:%p\n", &mov_r10_imm64 + 1, after);
+            __debugbreak();
```
bgianfo, line 422: The other tests seem to "return 1;" on failure to make the failure obvious.
Shall we follow that convention?

bgianfo (changes requested):

bgianfo (approved):

PR #118, opened 2020-08-12
Add IsWow64ProcessHelper function

zeffy:
I needed to compile Detours with VS 2005, which doesn't include `IsWow64Process` in its Windows SDK, so added a small wrapper function around it using `GetModuleHandleW` and `GetProcAddress`.

Used the [example on MSDN](https://docs.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process#examples) as a reference.

### DIFF
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -505,6 +505,9 @@ static BOOL IsWow64ProcessHelper(HANDLE hProcess,
                                  PBOOL Wow64Process)
 {
 #ifdef _X86_
+    UNREFERENCED_PARAMETER(hProcess);
+    UNREFERENCED_PARAMETER(Wow64Process);
+    
     return FALSE;
 #else
     if (Wow64Process == NULL) {
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -504,6 +504,9 @@ typedef BOOL(WINAPI *LPFN_ISWOW64PROCESS)(HANDLE, PBOOL);
 static BOOL IsWow64ProcessHelper(HANDLE hProcess,
                                  PBOOL Wow64Process)
 {
+#ifdef _X86_
+    return FALSE;
+#else
     if (Wow64Process == NULL) {
         return FALSE;
     }
@@ -526,6 +529,7 @@ static BOOL IsWow64ProcessHelper(HANDLE hProcess,
         return FALSE;
     }
     return pfnIsWow64Process(hProcess, Wow64Process);
+#endif
 }
 
 //////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


src/creatwth.cpp
```diff
@@ -499,6 +499,33 @@ static BOOL UpdateFrom32To64(HANDLE hProcess, HMODULE hModule, WORD machine,
...
+        return FALSE;
+    }
+
+    // IsWow64Process is not available on all supported versions of Windows.
+    // Use GetModuleHandle to get a handle to the DLL that contains the function
+    // and GetProcAddress to get a pointer to the function if available.
+
+    HMODULE hKernel32 = GetModuleHandleW(L"KERNEL32.DLL");
+    if (hKernel32 == NULL) {
+      return FALSE;
```
bgianfo, line 517: Should we add  DETOUR_TRACE(...); messages on the failure paths to aid folks who might be debugging?

src/creatwth.cpp
```diff
@@ -499,6 +499,33 @@ static BOOL UpdateFrom32To64(HANDLE hProcess, HMODULE hModule, WORD machine,
...
+                                 PBOOL Wow64Process)
+{
+    if (Wow64Process == NULL) {
+        return FALSE;
+    }
+
+    // IsWow64Process is not available on all supported versions of Windows.
+    // Use GetModuleHandle to get a handle to the DLL that contains the function
+    // and GetProcAddress to get a pointer to the function if available.
+
```
bgianfo, line 514: nit: Detours style seems to be to have either no empty line after a block comment, or a "//" line intead of an empty comment.

bgianfo:
Thanks for the patch @zeffy!
If you wouldn't mind addressing the minor comments I left, I'll be happy the merge.

src/creatwth.cpp
zeffy: I think under normal circumstances, this call to GetModuleHandleW can never fail. However I know there are some who use Detours (with modifications) in a context where only ntdll.dll is loaded, so I was being cautious. But yes, I agree and will add DETOUR_TRACE.

bgianfo: Agreed, I was more worried about the GetProcAddress() failure, if they are on a downlevel platform. I meant to call out that we should trace both.

zeffy: Ah, I see. Sure, I will add that as well.

bgianfo:
@dtarditi Any objections?

bgianfo:
@zeffy, sorry I think you also need to rebase ontop of latest master to get the new CI build to pass.

bgianfo:
Looks like this should also help with the changes that were needed to get detours to build, as mentioned in #91.

zeffy:
I've followed your suggestions, however calling it directly, even in non-X86 where it is guaranteed to exist on a system, defeats my original purpose of the PR to let Detours compile on VS 2005. 

As the function does not exist in the SDK headers (and I assume kernel32.lib as well, although I can't verify this right now), even when targeting X64.

bgianfo:
Thanks @zeffy !

PR #119, opened 2020-08-13
Define `_KERNEL32_` and `_USER32_` for the sake of users providing kernel32/user32 functionality for themselves.

jaykrell:
Define `_KERNEL32_` and `_USER32_` for the sake of users providing kernel32/user32 functionality for themselves.
This is not hypothetical. This fixed our own build breaks.

Mainstream users will suffer acceptably, one extra instruction on some calls, which LTCG optimizes away, if you actually care about performance. And function pointer comparisons kinda won't work, and nobody cares (each .dll gets own single instruction stub, instead of reading through the pointer to the main implementation).

### DIFF
--- a/src/detours.h
+++ b/src/detours.h
@@ -28,6 +28,15 @@
 #pragma warning(disable:4091) // empty typedef
 #endif
 
+// Suppress declspec(dllimport) for the sake of Detours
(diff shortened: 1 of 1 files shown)


bgianfo:
@jaykrell, can you rebase this on top of latest master so we can get build validation?

PR #121, opened 2020-08-19
Initial CI workflow for Detours

bgianfo:
Add a simple workflow that builds x86 and x64 build architectures on push to master, or to validate pull requests.
It currently builds with MSVC:
- x64
- x86

It also runs CodeQL analysis with the build, so we can get further static analysis from the CodeQL rule-sets.

In order to get the code building in on the latest Windows Kit SDK, I had restrict the use of GetThreadLocale/SetThreadlocale
under WINVER guard, as the SDK headers introduced the same check.

I've also included a commit to disable more logo banners, as that seems to be the convention in the rest of the build.

### DIFF
--- a/samples/traceapi/_win32.cpp
+++ b/samples/traceapi/_win32.cpp
@@ -4319,8 +4319,10 @@ BOOL (__stdcall * Real_GetThreadContext)(HANDLE a0,
 HDESK (__stdcall * Real_GetThreadDesktop)(DWORD a0)
     = GetThreadDesktop;
 
+#if(WINVER >= 0x0500)
 LCID (__stdcall * Real_GetThreadLocale)(void)
     = GetThreadLocale;
+#endif // (WINVER >= 0x0500)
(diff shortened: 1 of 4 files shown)


samples/traceapi/_win32.cpp
```diff
@@ -4319,8 +4319,10 @@ BOOL (__stdcall * Real_GetThreadContext)(HANDLE a0,
 HDESK (__stdcall * Real_GetThreadDesktop)(DWORD a0)
     = GetThreadDesktop;
 
+#if(WINVER >= 0x0500)
```
jaykrell, line 4322: No need for parens.

jaykrell: Also, can declare the function yourself.
Assuming the import libs have no such filtering.
I'm not sure which is best. Versioning is an unsolved mess (except for just staying at head like some large orgs..)

PR #122, opened 2020-08-20
early return in detour_alloc_region_from_hi and lo fail

danikdanik:
Return early in case the VirtualAlloc fails in detour_alloc_region_from_lo() and detour_alloc_region_from_hi() with ERROR_DYNAMIC_CODE_BLOCKED error, aka "The operation was blocked as the process prohibits dynamic code generation"


src/detours.cpp
```diff
@@ -18,6 +18,8 @@
 
 #define NOTHROW
 
+#define ERROR_DYNAMIC_CODE_BLOCKED       1655L // see winerror.h, as this error isn't found in some SDKs
```
bgianfo, line 21: Consider putting this in detours.h under the DETOURS_INTERNAL ifdef, incase it needs to be used elsewhere in the future.

bgianfo (changes requested):
Thanks for the patch @danikdanik!
If you don't mind fixing up the minor comment I left, this should be good to merge.

PR #129, opened 2020-08-28
Build: Catch incorrect format specifiers with /we4777 and fix issues.

bgianfo:
Most of these are using %d for a DWORD when it should be %ld.

### DIFF
--- a/samples/comeasy/wrotei.cpp
+++ b/samples/comeasy/wrotei.cpp
@@ -101,7 +101,7 @@ int WINAPI TimedEntryPoint(VOID)
     }
     else {
         printf("wrotei" DETOURS_STRINGIFY(DETOURS_BITS) ".dll:"
-               " Error detouring IStram::Wrote(): %d\n", error);
+               " Error detouring IStram::Wrote(): %ld\n", error);
     }
 
     printf("wrotei" DETOURS_STRINGIFY(DETOURS_BITS) ".dll:"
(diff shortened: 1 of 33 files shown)


src/Makefile
```diff
@@ -16,7 +16,7 @@ DETOURS_SOURCE_BROWSING = 0
 
 #######################/#######################################################
 ##
-CFLAGS=/nologo /W4 /WX /we4800 /Zi /MT /Gy /Gm- /Zl /Od
```
jaykrell, line 19: What happens with old compiler?

samples/comeasy/wrotei.cpp
```diff
@@ -101,7 +101,7 @@ int WINAPI TimedEntryPoint(VOID)
     }
     else {
         printf("wrotei" DETOURS_STRINGIFY(DETOURS_BITS) ".dll:"
-               " Error detouring IStram::Wrote(): %d\n", error);
+               " Error detouring IStram::Wrote(): %ld\n", error);
```
jaykrell, line 104: I've long been ambivalent about this sort of mismatch, because:
 a. They have the same underlying size.
 b. Signedness is still mixed up, but perhaps deliberately to affect the output format.
 c. Maybe some other system typedefs `DWORD` to be unsigned int instead, still 32bits, and the code is portable to that.
So the whole printf thing is broken really. Thinking more broadly, the format string is overly tied to the concrete type of input. Which could be a typedef and there could be a desired flexiblity.

It makes you wonder if iostreams should be used, so the code can become type-ignorant.

Or a macro to abstract, like PRINTF_FORMAT_DWORD, like is provided by stdint.h/inttypes.h/etc. for many types.

src/Makefile
bgianfo: I tested a warning introduced in VS 2019 on a VS 2017 version of MSVC, and it seemed to just ignore it.

PR #132, opened 2020-08-28
Members initialized in initialization list instead of ctor body

danikdanik:
Members should be initialized in the MIL


bgianfo:
👍 Thanks @danikdanik !

PR #138, opened 2020-09-04
add vs solution, it is safety because it only call nmake command, and it is friendly for edit, modify and debug code in Detours

sonyps5201314:
add vs solution, it is safety because it only call nmake command, and it is friendly for edit, modify and debug code in Detours

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,11 @@ include/
 obj.*
 *.ipdb
 *.iobj
+*.tlog
+*.log
+*.obj
+*.user
+*.recipe
+/bin.*
+*.vcxproj.FileListAbsolute.txt
+*.vcxprojAssemblyReference.cache
--- a/vc/Detours.sln
+++ b/vc/Detours.sln
@@ -1,0 +1,43 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29519.181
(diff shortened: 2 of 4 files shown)


sonyps5201314:
image
And it is very convenient,  because you can open every thing in project by visual studio.

bgianfo:
Thanks for the contribution!

I think this can go away when CMake support is finally merged with #48, as VS now supports CMake projects natively.
Until then, seems like a reasonable thing.
Thanks!

sonyps5201314:
CMake is avant-garde, but at the same time it also has a learning cost. It has a bunch of documents, and not every developer can edit it, so if the user needs to add a sample or compile parameters, participants may need to spend a lot of time to learn it, and it is not Microsoft's own product, so if you use it in such an important level of product, you are also doing free promotion for it, and Detours' nmake compilation script has been working stably for so many years, easy to understand, and compatible with basically all VS versions, so I don't think it is very urgent to migrate to CMake.

sylveon:
I'd argue that Makefiles have a bigger learning cost :)

Existing MS projects like wil already use cmake, so it's not like MS isn't already "doing promotion"

sonyps5201314:
The makefile of vc is like the BAT batch file of the Windows system. It is only by modifying various environment variables to set the compilation parameters to finally call the compiler component of vc, so it is very easy to understand. And in this project, you only need to slightly modify and adjust the compilation parameters, or add a sample, which can be drawn from the gourd.

PR #139, opened 2020-09-04
Merge code from detours 3.0 build 343

sonyps5201314:
Detours 4.0.1 looks like it's Detours 3.0 build_339 changed its name by comparing the code and looking at git commit logs, so the code in follow version need to be merged.


bgianfo:
@Jaykrell looks like this was your change originally? Care to review?

sonyps5201314:

PR #140, opened 2020-09-04
Fix can not createwith some protected exe

sonyps5201314:
fix can not use createwith api to start some unnormal exe, like chinese online、commercial、prorected game dnf.exe https://dnf.qq.com/, but this exe can start run by explorer.

### DIFF
--- a/src/uimports.cpp
+++ b/src/uimports.cpp
@@ -106,33 +106,35 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
         }
     }
 
-	///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-	//Added code
-	if (inh.IMPORT_DIRECTORY.VirtualAddress!=NULL && inh.IMPORT_DIRECTORY.Size==0)
-	{
-		//Don't worry about changing the PE file, because the load information of the original PE header has been saved and will be restored. The change here is just for the following code to work normally
-		PIMAGE_IMPORT_DESCRIPTOR pImageImport=(PIMAGE_IMPORT_DESCRIPTOR)(pbModule+inh.IMPORT_DIRECTORY.VirtualAddress);
-		IMAGE_IMPORT_DESCRIPTOR ImageImport;
+    if (inh.IMPORT_DIRECTORY.VirtualAddress != NULL && inh.IMPORT_DIRECTORY.Size == 0) {
+        
+        //Don't worry about changing the PE file, 
+        //because the load information of the original PE header has been saved and will be restored. 
+        //The change here is just for the following code to work normally
+        
+        PIMAGE_IMPORT_DESCRIPTOR pImageImport = (PIMAGE_IMPORT_DESCRIPTOR)(pbModule + inh.IMPORT_DIRECTORY.VirtualAddress);
+        IMAGE_IMPORT_DESCRIPTOR ImageImport;
 
-		if(!ReadProcessMemory(hProcess,pImageImport,&ImageImport,sizeof(ImageImport),NULL))
-		{
-			_Analysis_assume_(FALSE);
-			goto finish;
-		}
-		while(ImageImport.Name)
-		{
-			inh.IMPORT_DIRECTORY.Size+=sizeof(IMAGE_IMPORT_DESCRIPTOR);
-			if(!ReadProcessMemory(hProcess,pImageImport,&ImageImport,sizeof(ImageImport),NULL))
-			{
-				_Analysis_assume_(FALSE);
-				goto finish;
-			}
-			++pImageImport;
-		}
-		OutputDebugString(TEXT("[This PE file has an import table, but the import table size is marked as 0. This is an error.")
-			TEXT("Not repair will cause the launched program to not work normally, and its import table size has been automatically repaired for you! ! !]\r\n"));
-	}
-	///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+        if (!ReadProcessMemory(hProcess, pImageImport, &ImageImport, sizeof(ImageImport), NULL)) {
+            _Analysis_assume_(FALSE);
+            goto finish;
+        }
+        while (ImageImport.Name) {
+            inh.IMPORT_DIRECTORY.Size += sizeof(IMAGE_IMPORT_DESCRIPTOR);
+            if (!ReadProcessMemory(hProcess, pImageImport, &ImageImport, sizeof(ImageImport), NULL)) {
+                _Analysis_assume_(FALSE);
+                goto finish;
+            }
+            ++pImageImport;
+        }
+
+        DWORD dwLastError = GetLastError();
+        OutputDebugString(TEXT("[This PE file has an import table, but the import table size is marked as 0. This is an error.")
+            TEXT("If it is not repaired, the launched program will not work properly, Detours has automatically repaired its import table size for you! ! !]\r\n"));
+        if (GetLastError() != dwLastError) {
+            SetLastError(dwLastError);
+        }
+    }
 
     DETOUR_TRACE(("     Imports: %p..%p\n",
                   (DWORD_PTR)pbModule + inh.IMPORT_DIRECTORY.VirtualAddress,
--- a/src/uimports.cpp
+++ b/src/uimports.cpp
@@ -108,21 +108,21 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
 
     if (inh.IMPORT_DIRECTORY.VirtualAddress != NULL && inh.IMPORT_DIRECTORY.Size == 0) {
         
-        //Don't worry about changing the PE file, 
-        //because the load information of the original PE header has been saved and will be restored. 
-        //The change here is just for the following code to work normally
+        // Don't worry about changing the PE file, 
+        // because the load information of the original PE header has been saved and will be restored. 
+        // The change here is just for the following code to work normally
         
         PIMAGE_IMPORT_DESCRIPTOR pImageImport = (PIMAGE_IMPORT_DESCRIPTOR)(pbModule + inh.IMPORT_DIRECTORY.VirtualAddress);
         IMAGE_IMPORT_DESCRIPTOR ImageImport;
 
         if (!ReadProcessMemory(hProcess, pImageImport, &ImageImport, sizeof(ImageImport), NULL)) {
-            _Analysis_assume_(FALSE);
+            DETOUR_TRACE(("ReadProcessMemory failed: %u\n", GetLastError()));
             goto finish;
         }
         while (ImageImport.Name) {
             inh.IMPORT_DIRECTORY.Size += sizeof(IMAGE_IMPORT_DESCRIPTOR);
             if (!ReadProcessMemory(hProcess, pImageImport, &ImageImport, sizeof(ImageImport), NULL)) {
-                _Analysis_assume_(FALSE);
+                DETOUR_TRACE(("ReadProcessMemory failed: %u\n", GetLastError()));;
                 goto finish;
             }
             ++pImageImport;
--- a/src/uimports.cpp
+++ b/src/uimports.cpp
@@ -106,6 +106,34 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
         }
     }
 
+	///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+	//Added code
+	if (inh.IMPORT_DIRECTORY.VirtualAddress!=NULL && inh.IMPORT_DIRECTORY.Size==0)
+	{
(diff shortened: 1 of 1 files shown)


src/uimports.cpp
```diff
@@ -106,6 +106,34 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
...
+		{
+			inh.IMPORT_DIRECTORY.Size+=sizeof(IMAGE_IMPORT_DESCRIPTOR);
+			if(!ReadProcessMemory(hProcess,pImageImport,&ImageImport,sizeof(ImageImport),NULL))
+			{
+				_Analysis_assume_(FALSE);
+				goto finish;
+			}
+			++pImageImport;
+		}
+		OutputDebugString(TEXT("[This PE file has an import table, but the import table size is marked as 0. This is an error.")
```
bgianfo, line 132: We should use DETOUR_TRACE instead of OutputDebugStrring?

bgianfo (changes requested):
Thanks for the contribution!

It looks like you might have mistakenly included all of the other commits from your other PRs here?

src/uimports.cpp
sonyps5201314: DETOUR_TRACE can not ouput any message if no defined DETOUR_DEBUG, I use OutputDebugString to tell user to pay attention to this is a unnormal exe.

sonyps5201314:

src/uimports.cpp
sylveon: Should finish with a ] if it starts with [. Also OutputDebugString doesn't automatically append a new line, so you want to add `\n` at the end of the string.

src/uimports.cpp
```diff
@@ -106,6 +106,34 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
...
+	//Added code
+	if (inh.IMPORT_DIRECTORY.VirtualAddress!=NULL && inh.IMPORT_DIRECTORY.Size==0)
+	{
+		//Don't worry about changing the PE file, because the load information of the original PE header has been saved and will be restored. The change here is just for the following code to work normally
+		PIMAGE_IMPORT_DESCRIPTOR pImageImport=(PIMAGE_IMPORT_DESCRIPTOR)(pbModule+inh.IMPORT_DIRECTORY.VirtualAddress);
+		IMAGE_IMPORT_DESCRIPTOR ImageImport;
+
+		if(!ReadProcessMemory(hProcess,pImageImport,&ImageImport,sizeof(ImageImport),NULL))
+		{
+			_Analysis_assume_(FALSE);
```
sylveon, line 119: This tells the compiler it will never hit this point, so the goto right after will probably be completely elided.

sonyps5201314: image
from IDA decompiled code, the `goto `code can work fine.


fir

sylveon: Ah, confused it for __assume.

src/uimports.cpp
sonyps5201314: `\r\n` is in next line, this is a const string concat operation.

src/uimports.cpp
```diff
@@ -106,6 +106,34 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
         }
     }
 
+	///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+	//Added code
```
bgianfo, line 110: Please remove these comments which won't make sense when the PR is merged.

src/uimports.cpp
```diff
@@ -106,6 +106,34 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
...
+	if (inh.IMPORT_DIRECTORY.VirtualAddress!=NULL && inh.IMPORT_DIRECTORY.Size==0)
+	{
+		//Don't worry about changing the PE file, because the load information of the original PE header has been saved and will be restored. The change here is just for the following code to work normally
+		PIMAGE_IMPORT_DESCRIPTOR pImageImport=(PIMAGE_IMPORT_DESCRIPTOR)(pbModule+inh.IMPORT_DIRECTORY.VirtualAddress);
+		IMAGE_IMPORT_DESCRIPTOR ImageImport;
+
+		if(!ReadProcessMemory(hProcess,pImageImport,&ImageImport,sizeof(ImageImport),NULL))
```
bgianfo, line 117: nit; follow style `if (`, and braces are in the wrong place for this projects style.

src/uimports.cpp
```diff
@@ -106,6 +106,34 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
...
+			inh.IMPORT_DIRECTORY.Size+=sizeof(IMAGE_IMPORT_DESCRIPTOR);
+			if(!ReadProcessMemory(hProcess,pImageImport,&ImageImport,sizeof(ImageImport),NULL))
+			{
+				_Analysis_assume_(FALSE);
+				goto finish;
+			}
+			++pImageImport;
+		}
+		OutputDebugString(TEXT("[This PE file has an import table, but the import table size is marked as 0. This is an error.")
+			TEXT("Not repair will cause the launched program to not work normally, and its import table size has been automatically repaired for you! ! !]\r\n"));
```
bgianfo, line 133:
```cpp
TEXT("Not repair will cause the launched program to not work normally, and its import table size has been automatically repaired for you! ! !]\r\n");
```
I don't understand what this message is trying to tell the user, I don't think it makes sense how it's worded today. Can you re-word it to make it more clear?

src/uimports.cpp
bgianfo: > DETOUR_TRACE can not output any message if no define DETOUR_DEBUG, I use OutputDebugString to force output message for tell user to pay attention to this is a unnormal exe.

So the users is going to just happen to be running under a debugger? 
OutputDebugString only outputs a message if a debugger is attached.

It's not really detours job to emit messages as library.
I think it would make more sense to just return a rich
error code in LastError, and let the caller handle that case
explicitly.

src/uimports.cpp
bgianfo: space around boolean operators need spaces as well

src/uimports.cpp
```diff
@@ -106,6 +106,34 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
...
+			_Analysis_assume_(FALSE);
+			goto finish;
+		}
+		while(ImageImport.Name)
+		{
+			inh.IMPORT_DIRECTORY.Size+=sizeof(IMAGE_IMPORT_DESCRIPTOR);
+			if(!ReadProcessMemory(hProcess,pImageImport,&ImageImport,sizeof(ImageImport),NULL))
+			{
+				_Analysis_assume_(FALSE);
+				goto finish;
```
bgianfo, line 128: SetLastError in the failure paths?

src/uimports.cpp
```diff
@@ -106,6 +106,34 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
         }
     }
 
+	///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+	//Added code
+	if (inh.IMPORT_DIRECTORY.VirtualAddress!=NULL && inh.IMPORT_DIRECTORY.Size==0)
+	{
+		//Don't worry about changing the PE file, because the load information of the original PE header has been saved and will be restored. The change here is just for the following code to work normally
```
bgianfo, line 113: nit: wrap comments to a reasonable length?

src/uimports.cpp
```diff
@@ -106,6 +106,34 @@ static BOOL UPDATE_IMPORTS_XX(HANDLE hProcess,
...
+
+		if(!ReadProcessMemory(hProcess,pImageImport,&ImageImport,sizeof(ImageImport),NULL))
+		{
+			_Analysis_assume_(FALSE);
+			goto finish;
+		}
+		while(ImageImport.Name)
+		{
+			inh.IMPORT_DIRECTORY.Size+=sizeof(IMAGE_IMPORT_DESCRIPTOR);
+			if(!ReadProcessMemory(hProcess,pImageImport,&ImageImport,sizeof(ImageImport),NULL))
```
bgianfo, line 125: I think you could simplify this code by switching this to a do-while loop, that way the initial ReadProcessMemory doesn't need to be special cased.

src/uimports.cpp
sonyps5201314: [Dbgview ](https://docs.microsoft.com/en-us/sysinternals/downloads/debugview) and many debug tools can capture the debug output statements of the application without being debugged.
In fact, it doesn't matter whether the user notices this message, it's like an MFC running warning.
like 
```c++
		TRACE(traceAppMsg, 0, _T("Warning: calling DestroyWindow in CWnd::~CWnd; ")
		   _T("OnDestroy or PostNcDestroy in derived class will not be called.\n"));
```
in
```c++
CWnd::~CWnd()
{
	if (m_hWnd != NULL &&
		this != (CWnd*)&wndTop && this != (CWnd*)&wndBottom &&
		this != (CWnd*)&wndTopMost && this != (CWnd*)&wndNoTopMost)
	{
		TRACE(traceAppMsg, 0, _T("Warning: calling DestroyWindow in CWnd::~CWnd; ")
		   _T("OnDestroy or PostNcDestroy in derived class will not be called.\n"));
		DestroyWindow();
	}
```
The user cannot handle it because the only way to deal with it is to modify the PE file on the disk, but this file may contain a digital signature, and the modification may be illegal, and it only has a problem when it is launched through Detours' createwith API. The user uses Explorer. exe starts it, or directly calls the CreateProcess function to start it, there is no problem, these are enough to prove a defect of Detours.
The only improvement is that calling OutputDebugString in a non-debugging state may modify the value of LastError of the current thread, which has been fixed in the latest submission.

src/uimports.cpp
sonyps5201314: ReadProcessMemory will set LastError if  the function fails.

src/uimports.cpp
sonyps5201314: I don't quite understand your expression, and here `ImageImport.Name` is a judgment condition, so I think it is more appropriate to use a while-do loop.

src/uimports.cpp
sonyps5201314: The latest commit has resolved it.

src/uimports.cpp
sonyps5201314: The latest commit has resolved it.

src/uimports.cpp
sonyps5201314: The latest commit has resolved it.

src/uimports.cpp
sonyps5201314: The latest commit has resolved it.

bgianfo: Thanks!

src/uimports.cpp
bgianfo: Thanks!

src/uimports.cpp
sonyps5201314: the `_Analysis_assume_(FALSE);`  have been replaced with the `DETOUR_TRACE ` statements in latest commit.

src/uimports.cpp
halx99: The loop maybe better:
```cpp
for (;;) {
        if (!ReadProcessMemory(hProcess, pImageImport, &ImageImport, sizeof(ImageImport), NULL)) {
            DETOUR_TRACE(("ReadProcessMemory failed: %u\n", GetLastError()));;
            goto finish;
        }
        if (ImageImport.Name) {
            ++pImageImport;
            inh.IMPORT_DIRECTORY.Size += sizeof(IMAGE_IMPORT_DESCRIPTOR);
            continue;
        }
        break;
}
```

halx99: Updated

sonyps5201314: Although the code becomes able to get the correct result after you update and fix it, but at the same time it is difficult to read. In my original code, only run `inh.IMPORT_DIRECTORY.Size+=sizeof(IMAGE_IMPORT_DESCRIPTOR);` under the condition of `ImageImport.Name` , and you changed this code but first executed `inh.IMPORT_DIRECTORY.Size+=sizeof(IMAGE_IMPORT_DESCRIPTOR);` and ignored the conditions of `ImageImport.Name`. It seems that it is just to make the result correct, so I think it breaks the readability and there is no need to modify it.

halx99: why, it's easy to unstand, just calc size until the name is null

halx99: Update again, the latest code logical is equal to your original code

sonyps5201314: OK, now the code has been adjusted to be more in line with official recommendations.

bgianfo (approved):

bgianfo:
Thanks again @sonyps5201314 for all these fixes!

Issue #54, opened 2019-03-01
"Rejecting IJW module" exception injecting into .net


rob191:
When injecting any c++ dll via DetourCreateProcessWithDlls into a .net 2 executable that is configured with both .net 2 and .net 4 as supported runtimes, the clr throws exception code e0434352 (error "ERR: Rejecting IJW module built against v2.0.50727 because it could be loaded into another runtime in this process.") and the program terminates.  The same exe will load just fine with detours if the useLegacyV2RuntimeActivationPolicy option is set to true in the .net exe's .config file.  The clr function throwing the error and terminating the process is clr!PEFile::CheckForDisallowedInProcSxSLoadWorker and if that function is bypassed then the exe will load properly with detours.  Other methods of injecting the same dll using createremotethread or queueapc do not cause any issue and the exe loads fine.

I assume the issue has to do with the IAT method of injecting the dll but I am having trouble pinpointing the exact reason or modification from detours causing the issue to see if it can be corrected.

This is easily reproduced with the detours withdll exe, simple32.dll and an appropriate .net app to inject into.  I have attached a sample project.  If you build it in release mode and run "withdll.exe /d:simple32.dll DotNetAppTest.exe" you should reproduce the crash.

[DotNetAppTest.zip](https://github.com/Microsoft/Detours/files/2918112/DotNetAppTest.zip)

sonyps5201314:
you can try detors in my project to fix this problem,
https://developercommunityapi.westus.cloudapp.azure.com/storage/temp/11277-filetracker-bug-fix-src.zip,
I hope official will merge my modification to official repo.

rob191:
> you can try detours in my project to fix this problem,

I have built your version and it does appear to fix the issue.  I cannot read the comments in your code, there appears to be an encoding issue.  Would you mind explaining a little about what the issue with .net was and how you fixed it?  It appears that detours was not adding all of the appropriate information to the modified import table, is that correct?

rob191:
I was able to open up your changes with chinese encoding and translate your comments.  I will test the changes more but it does seem like they should be a welcome contribution to the trunk provided there are no issues.

jgdrg:
Hi. I'm seeing the same issue. The link posted by sonyps5201314  no longer works. Can anyone point me at the zip file or the code change that needs to be made?

rob191:
[...]

PR #141, opened 2020-09-04
Fix can not createwith all kinds of dot net exe

sonyps5201314:
fix can not use createwith api to start all kinds of .net exe, like Reflector.exe, it can be install by https://download.red-gate.com/ReflectorInstaller.exe

this commit revert "Improved Detours logic for detection of 32bit processes (#104)" , because there are some  bugs in that commit,
for example:
**1**. from detours wiki text:
`
Is Detours compatible with Windows 95, Windows 98, or Windows ME?
No. Detours is compatible only with the Windows NT family of operating systems: Windows NT, Windows XP, and Windows Server 2003, etc. Detours does not work on the Windows 9x family of operating systems because they have a primitive virtual memory system.
`
we known Detours compatible with  Windows 2000 and later system. but [IsWow64Process](https://docs.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process) is only available from Windows XP SP2, 
these code:
```c++
    if (!IsWow64ProcessHelper(GetCurrentProcess(), &bIs64BitOS)) {
        return FALSE;
    }
```
```c++
        if (!IsWow64ProcessHelper(hProcess, &bIs32BitProcess)) {
            return FALSE;
        }
```

are from "Improved Detours logic for detection of 32bit processes (#104)" commit,  and `IsWow64ProcessHelper ` will return FALSE when it can not find `IsWow64Process` function, 
```c++
    LPFN_ISWOW64PROCESS pfnIsWow64Process = (LPFN_ISWOW64PROCESS)GetProcAddress(
        hKernel32, "IsWow64Process");

    if (pfnIsWow64Process == NULL) {
        DETOUR_TRACE(("GetProcAddress failed: %d\n", GetLastError()));
        return FALSE;
    }
```
if `IsWow64ProcessHelper ` return FALSE, you will return FALSE too, and `DetourUpdateProcessWithDll ` will be failed.
so these code will make Detours  no longer compatible with all windows NT family.
**2**. "Improved Detours logic for detection of 32bit processes (#104)" commit delete follow code in `DetourUpdateProcessWithDll ` :
```c++
    BOOL bHas64BitDll = FALSE;
    BOOL bHas32BitExe = FALSE;
```

```c++
            if (inh.OptionalHeader.Magic == IMAGE_NT_OPTIONAL_HDR32_MAGIC
                && inh.FileHeader.Machine != 0) {

                bHas32BitExe = TRUE;
            }
```

```c++
        else {
            if (inh.OptionalHeader.Magic == IMAGE_NT_OPTIONAL_HDR64_MAGIC
                && inh.FileHeader.Machine != 0) {

                bHas64BitDll = TRUE;
            }
        }
```
it will make the follow code can not be compile success:
```c++
DETOUR_TRACE(("    32BitExe=%d 32BitProcess\n", bHas32BitExe, bIs32BitProcess));
```
Finally, my `Detour::Is64BitProcess` code Provides a more perfect solution, it have test on all different kinds of windows version from Windows 2000 to Winodws 10, include x86 and x64 architecture.


src/creatwth.cpp
```diff
@@ -557,46 +585,47 @@ BOOL WINAPI DetourUpdateProcessWithDll(_In_ HANDLE hProcess,
...
-        if (!IsWow64ProcessHelper(hProcess, &bIs32BitProcess)) {
-            return FALSE;
-        }
-    } else {
-        bIs32BitProcess = TRUE;
-    }
-
-    DETOUR_TRACE(("    32BitExe=%d 32BitProcess\n", bHas32BitExe, bIs32BitProcess));
+	/////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+	//Commented out code
```
bgianfo, line 610: Please don't include commented out code in PRs, thanks!

frerich:
There's a lot of stuff going on! My understanding is that there are two proposed improvements:

1. The line `    DETOUR_TRACE(("    32BitExe=%d 32BitProcess\n", bHas32BitExe, bIs32BitProcess));` should be removed since commit 81e6a5f79645b6fe385a2b72ee728962049dee6b removed the `bHas32BitExe` variable, causing builds with `DETOURS_DEBUG` defined to fail.

    I concur -- in fact, the `DETOUR_TRACE` line seems subious since there is just one format specifier (`%d`) but two arguments. Did this ever do the right thing? Assuming that debug output is useful here, an alternative fix would be to _adjust_ (instead of _removing_) the statement to e.g.

    ```c
    DETOUR_TRACE(("    64BitOS=%d 32BitProcess=%d\n", bIs64BitOS, bIs32BitProcess));
    ```

    ...i.e. printing the value of a different (closely related) variable.

2. The function `IsWow64ProcessHelper` should not be used since it returns `FALSE` on operating systems which are supported according to the Wiki, but which do not feature the `IsWow64Process` function. I don't understand this argument yet: in case an operating system does not support the `IsWow64Process` function, isn't it safe to assume that the system does not feature a WoW64 subsystem and thus, trivially, no process can be running under WoW64? In that case, returning `FALSE` seems plausible to me.

    Or is the argument that the definition for `bIs64BitOS` is incorrect since there are supported versions of Windows which do use 64bit, but which do not support `IsWow64process`? I suppose that's plausible, but I'm not aware of any such Windows version - can anybody else shed some light?

sonyps5201314:
@frerich
**1**.
DETOUR_TRACE((" 32BitExe=%d 32BitProcess\n", bHas32BitExe, bIs32BitProcess)); should be a clerical error, so `32BitProcess` should be change to `32BitProcess=%d` to use the second variable parameter `bIs32BitProcess`, in my commit, it have done.
**2**.
you can know which systems support `IsWow64Process` from MSDN: https://docs.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process
`
Minimum supported client | Windows Vista, Windows XP with SP2 [desktop apps \| UWP apps]
`
it is authoritative.
**3**. if you have look the code in the leaked Windows 2000 source code, you will known that Windows 2000 is a exampled system that support Wow64 but have not `IsWow64Process` function,
image
and this facts do not accord with your imagination, 
`I don't understand this argument yet: in case an operating system does not support the IsWow64Process function, isn't it safe to assume that the system does not feature a WoW64 subsystem and thus, trivially, no process can be running under WoW64? I`
saied by you.
**4.**
you can test everything by install all kinds of Windows by VMware Workstation ,to verify your hypothesis , instead of being confused.

sonyps5201314:

src/creatwth.cpp
sonyps5201314: @bgianfo, I have removed the commented out code

frerich:
@sonyps5201314 Thanks for elaborating!

1. From what I've gathered, the `DETOUR_TRACE` statement in question never worked: it used to lack a format specifier for the second argument ever since the source code of Detours 4.0 was imported November 2006. I suppose that shows how often the `DETOUR_DEBUG` flag is set. :-) I think fixing this makes perfect sense.

2. If I understood you right, the issue with using `IsWow64Process` to detect 64 bit operating systems is that there is at least one system (Windows 2000 64bit?) which _does_ support the WoW64 subsystem but which does _not_ feature the `IsWow64Process` function. Thus, if Detours is compiled for 32bit, the code might incorrectly decide that the operating system is 32bit, too.

     According to the blog post [Microsoft Unveils Plans for 64-Bit Windows Platform
](https://news.microsoft.com/2001/05/23/microsoft-unveils-plans-for-64-bit-windows-platform/), the first Windows versions which featured a 64bit architecture was Windows XP. The [Wikipedia page for Windows 2000](https://en.wikipedia.org/wiki/Windows_2000) (I fully acknowledge that this may not be an authoritative source) elaborates:

    > Windows was Windows XP 64-Bit Edition, released alongside the 32-bit editions of Windows XP on October 25, 2001,[35] followed by the server versions Windows Datacenter Server Limited Edition and later Windows Advanced Server Limited Edition, which were based on the pre-release Windows Server 2003 (then known as Windows .NET Server) codebase.

    My impression is that it's not Windows 2000 which is the tricky case, but it's Windows XP prior to SP2. I suspect that there may have been installations of Windows XP 64-Bit Edition which do _not_ have SP2 installed -- and the [`IsWow64Process` API documentation](https://docs.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process) mentions that this function is only available in SP2 or later.

    Now, with all that being said, I'm not opposed to supporting those Windows XP (or maybe even Windows 2000, as you argue) installations. However, I think it could be argued that the code, as it is, is simpler than it used to be (it no longer parses PE headers) which is a reasonable improvement in exchange for dropping support for Windows XP prior to SP 2. I'm not dogmatic about this though (we don't have to support customers using such old platforms, so the status quo is good enoug hfor me).

frerich:
For what it's worth, in case the Detours project should continue to support Windows NT, Windows XP (at least prior to SP2), and Windows Server 2003, it would probably be worthwhile to consider setting up some CI for this since I suspect that many external contributors might not take those platforms into account.

sonyps5201314:
Supporting these old platforms will not increase the amount of code. I don't think it is necessary to spend too much time on this issue. The core foundation of Detours is VirtualAlloc and VirtualProtect, and it has PAGE_EXECUTE permissions. As long as these conditions can be guaranteed.
My Is64BitOS and Is64BitProcess functions have served my various programs for many years, and they have also been tested in different Windows 2000 and later official release versions of Windows systems, so I believe they are stable and universal, your new appended judgment the logic of 64-bit systems and 32-bit processes does have some incomplete considerations, which is also a fact, so I think we can consider using a more reliable and comprehensive solution.

sonyps5201314:
from [Windows 2000 wiki page](https://en.wikipedia.org/wiki/Windows_2000)， 
`
Microsoft planned to release a 64-bit version of Windows 2000, which would run on 64-bit Intel Itanium microprocessors, in 2000.[33][34] However, the first officially released 64-bit version of Windows was Windows XP 64-Bit Edition, released alongside the 32-bit editions of Windows XP on October 25, 2001,[35] followed by the server versions Windows Datacenter Server Limited Edition and later Windows Advanced Server Limited Edition, which were based on the pre-release Windows Server 2003 (then known as Windows .NET Server) codebase.[36][37] These editions were released in 2002, were shortly available through the OEM channel and then were superseded by the final versions of Server 2003.[37]
`
we can know  `Microsoft planned to release a 64-bit version of Windows 2000, which would run on 64-bit Intel Itanium microprocessors, in 2000`，Explain the capabilities of Wow64 that Windows 2000 already has, the first picture I posted above is also a confirmation.
`However, the first officially released 64-bit version of Windows was Windows XP 64-Bit Edition, released alongside the 32-bit editions of Windows XP on October 25, 2001,` Explain that the first 64-bit official release version is XP, released in 2001, then this version definitely does not have the IsWow64Process API

frerich:
@sonyps5201314 It would be great if you could address the question in https://github.com/microsoft/Detours/pull/104#issuecomment-682356544 -- what does your definition of `Is64BitOS` return when running 32bit Windows on a 64bit CPU?

sonyps5201314:
> For what it's worth, in case the Detours project should continue to support Windows NT, Windows XP (at least prior to SP2), and Windows Server 2003, it would probably be worthwhile to consider setting up some CI for this since I suspect that many external contributors might not take those platforms into account.

If it's worthless,  official people would not let you rebase your code by  `IsWow64ProcessHelper `

sonyps5201314:
> @sonyps5201314 It would be great if you could address the question in [#104 (comment)](https://github.com/microsoft/Detours/pull/104#issuecomment-682356544) -- what does your definition of `Is64BitOS` return when running 32bit Windows on a 64bit CPU?

The theoretical result should be `FALSE`,  and the actual result is also `FALSE ` as you see below picture.
image

you can test by your test by complile below by any vc version from 6.0 as a console app.
```c++


#include <stdio.h>
#include <windows.h>

#ifndef PROCESSOR_ARCHITECTURE_AMD64
#define PROCESSOR_ARCHITECTURE_AMD64            9
#endif

#ifndef PROCESSOR_ARCHITECTURE_ARM64
#define PROCESSOR_ARCHITECTURE_ARM64            12
#endif

BOOL Is64BitOS()
{
	BOOL bRet = FALSE;
	HMODULE hModule = GetModuleHandle(TEXT("kernel32.dll"));
	if (!hModule)
	{
		return bRet;
	}
	VOID(WINAPI * _GetNativeSystemInfo)(OUT LPSYSTEM_INFO lpSystemInfo) = (void(__stdcall*)(LPSYSTEM_INFO))GetProcAddress(hModule, "GetNativeSystemInfo");
	if (!_GetNativeSystemInfo)
	{
		return bRet;
	}

	SYSTEM_INFO si;
	_GetNativeSystemInfo(&si);
	if (si.wProcessorArchitecture == PROCESSOR_ARCHITECTURE_AMD64 || si.wProcessorArchitecture == PROCESSOR_ARCHITECTURE_ARM64 ||
		si.wProcessorArchitecture == PROCESSOR_ARCHITECTURE_IA64 || si.wProcessorArchitecture == PROCESSOR_ARCHITECTURE_ALPHA64)
	{
		bRet = TRUE;
	}
	return bRet;
}

int main()
{
	BOOL bIs64BitOS = Is64BitOS();
	printf("bIs64BitOS=%d", bIs64BitOS);
	printf("\n", bIs64BitOS);
	getchar();

	return 0;
}

```
In my 64 bit Windows 8.1, it is `TRUE `as expected.
image

The reason you are puzzled is because you don’t understand the function of [GetNativeSystemInfo](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getnativesysteminfo).
According to MSDN，It only return the real hardware instruction architecture, even if the current process is running under WOW64.
And, you have another doubt,  Whether 64-bit CPU can install and run 32-bit system, you may use the computer late, I can tell you for sure, my first computer is Lenovo C466A, its processor is Intel Pentium T2390, it is a x64 CPU, I installed 32-bit Windows XP on it, and used it for many years until early 2014.

frerich:
@sonyps5201314 Thanks for the explanation. Just to clarify my understanding:

* You identified that the code currently doesn't compile anymore in case `DETOUR_DEBUG` is defined.  This is because the line `DETOUR_TRACE(("    32BitExe=%d 32BitProcess\n", bHas32BitExe, bIs32BitProcess));` -- which has always been defective ever since the code got imported into Git due to the missing formatting flag for the second argument -- references the variable `bHas32BitExe` which got removed in commit 81e6a5f79645b6fe385a2b72ee728962049dee6b

* You propose to improve the test for whether the operating system is 32bit or 64bit by using `GetNativeSystemInfo` instead of `IsWow64Process`. As far as I can see, the [`GetNativeSystemInfo`](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getnativesysteminfo) function is available since Windows XP whereas `IsWow64Process` is available only in Windows XP SP2 and later.

Is this accurate?

If so, can you clean this PR such that only those two aspects are addressed? As it is, there seem to be a lot of unrelated changes which make it a bit hard to see what's going on, e.g. in [modules.cpp:872](https://github.com/microsoft/Detours/pull/141/commits/ec9696d97d0e174cb583f98ab0b238c5cec06795#diff-eb51b101c32a1e2424caceb0d81b1aa9L872). Also, the `bHas32BitExe` variable is reintroduced but it doesn't seem to be used anywhere -- except for the debug output (but then, why bother including it in debug output if the value isn't used anywhere?).

sonyps5201314:
1.
In the official original code (including the official final commercial version 3.0 build 343), the logical judgment of bIs32BitProcess is indeed problematic, but it has now been corrected and it is perfect. It is because of defects that the corresponding formatting parameters are not given and it is just a guess of you. If this sentence DETOUR_TRACE is meaningless, it is estimated that the official has deleted it a long time ago. Only the original developer of this code knows the reason. . So I think that keeping this code is a respect for the original author, and we cannot determine the original author's intentions and expressions, so we cannot decide to remove it. Detours is great. It uses such a small code to implement x86, x64, ia64, arm, arm64 hooks. There is no similar software on the market that can support such a comprehensive and stable instruction set, but Microsoft has published it in 2002. In the future, it may also support linux/unix to complete all operating systems, expand Microsoft’s influence and contribution to the world, and win more praise. It has served various commercial projects for many years. Selling it at a high price is enough to show its technical content. This is Microsoft's official high affirmation of the value of this project. Including NVIDIA is also a customer of this software. You can find it in NVIDIA's graphics driver. Therefore, when we modify these codes, we should think twice and stay in awe, so as not to make it unstable. This is related to Microsoft's reputation.
2.
as you known,
`However, the first officially released 64-bit version of Windows was Windows XP 64-Bit Edition, released alongside the 32-bit editions of Windows XP on October 25, 2001,`
from https://en.wikipedia.org/wiki/Windows_2000
```
Windows XP 64-Bit Edition
"Windows XP 64-Bit Edition" redirects here. It is not to be confused with Windows XP Professional x64 Edition.
Windows XP 64-Bit Edition was designed to run on Intel Itanium family of microprocessors in their native IA-64 mode.

Two versions of Windows XP 64-Bit Edition were released:

Windows XP 64-Bit Edition for Itanium systems, Version 2002 – Based on Windows XP codebase, was released simultaneously alongside the 32-Bit version of Windows XP on October 25, 2001.[35]
Windows XP 64-Bit Edition, Version 2003 – Based on Windows Server 2003 codebase (which added support for the Itanium 2 processor), was released on March 28, 2003.[36]
```
from https://en.wikipedia.org/wiki/Windows_XP_editions#Windows_XP_64-Bit_Edition
XP IA64, released in 2001, was the first 64-bit system officially released by Microsoft, and GetNativeSystemInfo was available since XP, so does it mean that it is completely fine to use GetNativeSystemInfo to determine that the system is 64-bit.

3.my #141's purpose is 'Fix can not createwith all kinds of dot net exe', and not only 'Determine whether the process is 32-bit',
and 'Determine whether the process is 32-bit' is only a part of this task. [modules.cpp:872](https://github.com/microsoft/Detours/pull/141/commits/ec9696d97d0e174cb583f98ab0b238c5cec06795#diff-eb51b101c32a1e2424caceb0d81b1aa9L872) and so on，they are the rest  of this task. If not do these, can not fix all kinds of .net app, like the app build from https://github.com/Microsoft/Detours/files/2918112/DotNetAppTest.zip in #54 , you can not understand it, if you don`t known double process debug and debug in ntdll. This involves the core issues of the PE loader and the details of the .net clr runtime，but you can understand a little bit from the comments modified in my commit.
That's all, thanks for your attention.

src/creatwth.cpp
```diff
@@ -557,46 +587,30 @@ BOOL WINAPI DetourUpdateProcessWithDll(_In_ HANDLE hProcess,
...
+
+                bHas32BitExe = TRUE;
+            }
             DETOUR_TRACE(("%p  Found EXE\n", hLast));
         }
+        else {
+            if (inh.OptionalHeader.Magic == IMAGE_NT_OPTIONAL_HDR64_MAGIC
+                && inh.FileHeader.Machine != 0) {
+
+                bHas64BitDll = TRUE;
```
bgianfo, line 601: Value is never read?

src/creatwth.cpp
```diff
@@ -714,6 +728,10 @@ BOOL WINAPI DetourUpdateProcessWithDllEx(_In_ HANDLE hProcess,
         DETOUR_CLR_HEADER clr;
         CopyMemory(&clr, &der.clr, sizeof(clr));
         clr.Flags &= ~COMIMAGE_FLAGS_ILONLY;    // Clear the IL_ONLY flag.
+        //The above sentence clears the IL_ONLY flag, 
+        //so we need to restore the IL_ONLY flag after the IAT is restored after starting the process completion
```
bgianfo, line 732: >// [](start = 8, length = 2)

nit: all comments should have an empty space after comment characters, that's the style.

src/creatwth.cpp
```diff
@@ -499,37 +508,57 @@ static BOOL UpdateFrom32To64(HANDLE hProcess, HMODULE hModule, WORD machine,
 }
 #endif // DETOURS_64BIT
 
-typedef BOOL(WINAPI *LPFN_ISWOW64PROCESS)(HANDLE, PBOOL);
+namespace Detour {
```
bgianfo, line 511: >Detour [](start = 10, length = 6)

Detours? 

Also why does this need a name space, these should all be static functions?

src/creatwth.cpp
```diff
@@ -499,37 +508,57 @@ static BOOL UpdateFrom32To64(HANDLE hProcess, HMODULE hModule, WORD machine,
...
+            return bRet;
+        }
 
-    if (pfnIsWow64Process == NULL) {
-        DETOUR_TRACE(("GetProcAddress failed: %d\n", GetLastError()));
-        return FALSE;
+        SYSTEM_INFO si;
+        _GetNativeSystemInfo(&si);
+        if (si.wProcessorArchitecture == PROCESSOR_ARCHITECTURE_AMD64 || si.wProcessorArchitecture == PROCESSOR_ARCHITECTURE_ARM64 ||
+            si.wProcessorArchitecture == PROCESSOR_ARCHITECTURE_IA64 || si.wProcessorArchitecture == PROCESSOR_ARCHITECTURE_ALPHA64) {
```
bgianfo, line 536: >| si.wProcessorArchitecture == PROCESSOR_ARCHITECTURE_ALPHA64 [](start = 70, length = 61)

Detours doesn't support ALPHA, no point in checking

src/creatwth.cpp
sonyps5201314: this above code is just a revert of 81e6a5f79645b6fe385a2b72ee728962049dee6b, It has always existed in the previous version. It should have been written by one of your microsoft technical experts. Keeping it is only as a respect for the original author. Keeping it will not affect it. Maybe it will be used in the future.

src/creatwth.cpp
sonyps5201314: Is64BitOS exists as a general-purpose function. It is ported from my private library. Detours users can also copy and reuse it in other codes. It is judged that all current 64-bit Windows operating systems conform to the meaning of the function name, and here, although Detours does not support the ALPHA/ALPHA64 architecture, the judgment of this condition does not affect it. Because it just tells the caller whether the current system is 64-bit. Whether the current process is 32-bit is also determined by the Is64BitProcess function.

src/creatwth.cpp
sonyps5201314: The latest commit has resolved it.

src/creatwth.cpp
sonyps5201314: The latest commit has resolved it.

src/creatwth.cpp
sonyps5201314: image
from the IDA decompiled code, we can known that, my guess is correct, the smart VC compiler can really optimize out the code that is not used in the function.

frerich:
I still think that it makes sense to

1. Reduce this PR to the absolute minimum required to fix the issues described in the initial comment (i.e. making Detours compile in case `DETOUR_DEBUG` is defined and defining `bIs64BitOS` in terms of `GetNativeSystemInfo` instead of `IsWow64Helper`)

2. Split the resulting diff into two PRs, one for fixing the `DETOUR_DEBUG` compile issue (I suppose that PR will be a oneliner) and one for adjusting the logic for identifying 64bit OS.

sonyps5201314:
The purpose of this PR is to fix the createwith startup problem of all types of .net programs. Using a PR allows other users to understand how to fix it, and how to fix it by modifying the wrong places. And dividing into two PRs as you said is more like a fix you submitted before, which does not meet the purpose of this PR. I also explained before that in order to achieve this goal, it is not just a modification of the code in this place. Users can also open this PR to understand the details of our discussion.

frerich:
@bgianfo Thanks for bringing this PR to my attention! I finished reviewing this code and added some questions and concerns as comments to this PR.

bgianfo:
Thanks @frerich for taking a look.

bgianfo:
@sonyps5201314 / @frerich can we please seperate the revert of the #104 commit and fix so the history is clean?

The current proposed PR muddles them togeather.
I would like to get this merged and included in the next release, thank you for your patience!

frerich:
@bgianfo Thanks for bringing this up again! [It appears](https://github.com/microsoft/Detours/pull/141#issuecomment-687643082) that I understood this PR to improve two aspects of the source code, one of which being the portability of the code. PR #104 introduced the usage of the `IsWow64Helper` function and assumes that systems lacking this function cannot possibly run 64bit applications. However, it appears that Windows XP prior to SP2 _does_ support 64bit application but it _does not_ feature `IsWow64Helper`. Does that seem accurate?

If so, I find it hard to come up with a tested PR since I lack access to Windows XP installations which do not have SP2 installed. I could do a _blind patch_ at best (i.e. a patch which I never tried myself), but I'd rather not do that. :-)

@sonyps5201314 [quoted](https://github.com/microsoft/Detours/pull/141#issue-479283477) the Detours wiki, which pointed out that "Detours is compatible only with the Windows NT family of operating systems: Windows NT, Windows XP, and Windows Server 2003". Is this still accurate? If so, I suspect that all contributors to the project would benefit greatly from a CI setup which makes sure that all PRs are exercised (at least built) on all supported platforms. I'm afraid that otherwise, such portability issues might creep in rather sooner than later.

sonyps5201314:
> @sonyps5201314 / @frerich can we please seperate the revert of the #104 commit and fix so the history is clean?
> 
> The current proposed PR muddles them togeather.
> I would like to get this merged and included in the next release, thank you for your patience!

YES，however,  the first submission of this PR of mine has deleted the relevant code of @frerich because it conflicts with mine. Therefore, there should be no confusion.

sonyps5201314:
The merge conflict has been fixed.

src/creatwth.cpp
```diff
@@ -501,35 +510,53 @@ static BOOL UpdateFrom32To64(HANDLE hProcess, HMODULE hModule, WORD machine,
...
+#define PROCESSOR_ARCHITECTURE_ARM64            12
+#endif
 
-    if (pfnIsWow64Process == NULL) {
-        DETOUR_TRACE(("GetProcAddress failed: %lx\n", GetLastError()));
-        return FALSE;
+static BOOL Is64BitOS()
+{
+    BOOL bRet = FALSE;
+    HMODULE hModule = GetModuleHandle(TEXT("kernel32.dll"));
```
sylveon, line 524: Suggestion: directly call GetModuleHandleW here

sonyps5201314: No effect, because `TEXT` is the `winnt.h` header file from the Windows SDK, not the` tchar.h` of the CRT

sylveon: And Detours' makefile never defines UNICODE so this goes through WinAPI's internal code page translation when we could just hardcode the wide string instead.

sylveon: Examples in the existing code:
https://github.com/microsoft/Detours/blob/8cbb9e2b822d96d534400da915a5dc258ab65a9a/src/modules.cpp#L58
https://github.com/microsoft/Detours/blob/8cbb9e2b822d96d534400da915a5dc258ab65a9a/src/modules.cpp#L343

sonyps5201314: Okay, follow your Suggestion.

PR #142, opened 2020-09-04
Fix can not createwith exe which path contain special characters or unicode languages

sonyps5201314:
fix can not use createwith api to start exes which path contain special characters or unicode languages like Chinese

### DIFF
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -1310,8 +1310,7 @@ BOOL WINAPI DetourProcessViaHelperDllsW(_In_ DWORD dwTargetPid,
     //so we can't use "%hs" in format string, because the dll that contain this code would inject to any process, even not call _tsetlocale(LC_ALL,_T(".ACP")) before
     WCHAR szDllName[MAX_PATH];
     int cchWrittenWideChar = MultiByteToWideChar(CP_ACP, 0, &helper->rDlls[0], -1, szDllName, ARRAYSIZE(szDllName));
-    if (cchWrittenWideChar >= ARRAYSIZE(szDllName) || cchWrittenWideChar <= 0)
-    {
+    if (cchWrittenWideChar >= ARRAYSIZE(szDllName) || cchWrittenWideChar <= 0) {
         goto Cleanup;
     }
     hr = StringCchPrintfW(szCommand, ARRAYSIZE(szCommand),
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -1208,8 +1208,10 @@ BOOL WINAPI DetourProcessViaHelperDllsA(_In_ DWORD dwTargetPid,
         goto Cleanup;
     }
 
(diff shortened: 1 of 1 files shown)


bgianfo (changes requested):
Thanks for the contribution!

It looks like you might have mistakenly included  other commits from your other PRs here?

It looks like only a3e8ba1c29b0ed258d1bff1761b2cf163252c1ef is applicable to this PR?

sonyps5201314:
@bgianfo, I have rebased to only include this commit, and the commented out code have be removed.

src/creatwth.cpp
```diff
@@ -1304,8 +1306,12 @@ BOOL WINAPI DetourProcessViaHelperDllsW(_In_ DWORD dwTargetPid,
         goto Cleanup;
     }
 
-    hr = StringCchPrintfW(szCommand, ARRAYSIZE(szCommand),
-                          L"rundll32.exe \"%hs\",#1", &helper->rDlls[0]);
+	//for east asia languages and so on, like Chinese, print format with "%hs" can not work fine before user call _tsetlocale(LC_ALL,_T(".ACP"));
+    //so we can't use "%hs" in format string, because the dll that contain this code would inject to any process, even not call _tsetlocale(LC_ALL,_T(".ACP")) before
+	WCHAR szDllName[MAX_PATH];
+	MultiByteToWideChar(CP_ACP, 0, &helper->rDlls[0], -1, szDllName, ARRAYSIZE(szDllName));
```
bgianfo, line 1312: We need to check the return code of MultiByteToWideChar for error?

src/creatwth.cpp
```diff
@@ -1304,8 +1306,12 @@ BOOL WINAPI DetourProcessViaHelperDllsW(_In_ DWORD dwTargetPid,
         goto Cleanup;
     }
 
-    hr = StringCchPrintfW(szCommand, ARRAYSIZE(szCommand),
-                          L"rundll32.exe \"%hs\",#1", &helper->rDlls[0]);
+	//for east asia languages and so on, like Chinese, print format with "%hs" can not work fine before user call _tsetlocale(LC_ALL,_T(".ACP"));
+    //so we can't use "%hs" in format string, because the dll that contain this code would inject to any process, even not call _tsetlocale(LC_ALL,_T(".ACP")) before
```
bgianfo, line 1310: nit pick: Can we fix the comment here and above to be consistent with the comment style with the rest of the code?

bgianfo: indention of StringChhPrintf seems to be off by one, in both places you changed as well.

sonyps5201314: The latest commit has resolved it.

src/creatwth.cpp
sonyps5201314: The latest commit has resolved it.

bgianfo (approved):

bgianfo:
Thanks again for the contribution @sonyps5201314 !

PR #143, opened 2020-09-04
Fix: Delete payload after finish associated task

sonyps5201314:
Delete the payload after finish associated task, and need change the DETOUR_EXE_RESTORE_GUID's value for compatible with these dlls that compiled by old version Detours

for example,  create a pair of dlls named dll_createwith_mysocks_32/64.dll, their function is hooked CreateProcess to create any new process with it. 
a 32bit exe(named exe_socks) use dll_createwith_mysocks_32.dll by PE import table. so it start any process will with dll_createwith_mysocks_32/64.dll.
now it start a 32bit exe (named exe_vs), but exe_vs will load a dll (named dll_createwith_myluainject_32/64.dll) that build with detours and their function is hooked CreateProcess to create any new process with it.
so if  exe_vs start a 64bit exe, DetourCreateProcessXXX api will only restore the first IAT which modifed by dll_createwith_mysocks_32/64.dll, and dll_createwith_myluainject_32/64.dll modified IAT will not restore. because they create payload with the same GUID  DETOUR_EXE_RESTORE_GUID.  and Detour will do restore with the first founded payload twice.
so make dll_createwith_mysocks_32/64.dll can not work ok.
so we need delete the payload after the associated task finish immediately. if do like this, the payload with GUID  DETOUR_EXE_RESTORE_GUID can be used by the next dll that complied by Detours.

And for compatible with these dlls that compiled by old version Detours which we don`t have source code to recompile these.
so we need change  DETOUR_EXE_RESTORE_GUID's value to a new value, so even these dlls can not delete the payload, because we only search the new DETOUR_EXE_RESTORE_GUID's value's payload. so the old playload things will not executed by us.

### DIFF
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -11,7 +11,6 @@
 #define DETOURS_INTERNAL
 #include "detours.h"
 #include <stddef.h>
-#include <assert.h>
 
 #if DETOURS_VERSION != 0x4c0c1   // 0xMAJORcMINORcPATCH
 #error detours.h version mismatch
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -11,6 +11,7 @@
 #define DETOURS_INTERNAL
 #include "detours.h"
 #include <stddef.h>
+#include <assert.h>
 
 #if DETOURS_VERSION != 0x4c0c1   // 0xMAJORcMINORcPATCH
 #error detours.h version mismatch
@@ -1018,6 +1019,18 @@ VOID CALLBACK DetourFinishHelperProcess(_In_ HWND,
         delete[] rlpDlls;
         rlpDlls = NULL;
     }
+
+	//Delete the payload after execution to release the memory occupied by it
+	if (s_pHelper != NULL)
+	{
+		HMODULE hModule = DetourGetContainingModule(s_pHelper);
+		assert(hModule != NULL);
+		if (hModule != NULL)
+		{
+			VirtualFree(hModule, 0, MEM_RELEASE);
+			hModule = NULL;
+		}
+	}
 
     ExitProcess(Result);
 }
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -1019,7 +1019,7 @@ VOID CALLBACK DetourFinishHelperProcess(_In_ HWND,
         rlpDlls = NULL;
     }
 
-    //Delete the payload after execution to release the memory occupied by it
+    // Delete the payload after execution to release the memory occupied by it
     if (s_pHelper != NULL) {
         DetourFreePayload(s_pHelper);
     }
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -1020,8 +1020,7 @@ VOID CALLBACK DetourFinishHelperProcess(_In_ HWND,
     }
 
     //Delete the payload after execution to release the memory occupied by it
-    if (s_pHelper != NULL)
-    {
+    if (s_pHelper != NULL) {
         DetourFreePayload(s_pHelper);
     }
 
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -1023,13 +1023,7 @@ VOID CALLBACK DetourFinishHelperProcess(_In_ HWND,
 	//Delete the payload after execution to release the memory occupied by it
 	if (s_pHelper != NULL)
 	{
-		HMODULE hModule = DetourGetContainingModule(s_pHelper);
-		assert(hModule != NULL);
-		if (hModule != NULL)
-		{
-			VirtualFree(hModule, 0, MEM_RELEASE);
-			hModule = NULL;
-		}
+        DetourFreePayload(s_pHelper);
 	}
 
     ExitProcess(Result);
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -1020,11 +1020,11 @@ VOID CALLBACK DetourFinishHelperProcess(_In_ HWND,
         rlpDlls = NULL;
     }
 
-	//Delete the payload after execution to release the memory occupied by it
-	if (s_pHelper != NULL)
-	{
+    //Delete the payload after execution to release the memory occupied by it
+    if (s_pHelper != NULL)
+    {
         DetourFreePayload(s_pHelper);
-	}
+    }
 
     ExitProcess(Result);
 }
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -887,6 +887,11 @@ BOOL WINAPI DetourCopyPayloadToProcess(_In_ HANDLE hProcess,
         return FALSE;
     }
 
+    // As you can see in the following code,
+    // the memory layout of the payload range "[pbBase, pbBase+cbTotal]" is a PE executable file,
+    // so DetourFreePayload can use "DetourGetContainingModule(Payload pointer)" to get the above "pbBase" pointer,
+    // pbBase: the memory block allocated by VirtualAllocEx will be released in DetourFreePayload by VirtualFree.
+
     PBYTE pbTarget = pbBase;
     IMAGE_DOS_HEADER idh;
     IMAGE_NT_HEADERS inh;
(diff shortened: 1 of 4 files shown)


sylveon:
Could this perhaps be factored out into a public DetoursFreePayload method?

sonyps5201314:
> Could this perhaps be factored out into a public DetoursFreePayload method?

Yes, the official people have rights to append a new api.  we do this only for solve problems as soon as possible, and have no time to wait the official people's decision.

sonyps5201314:

src/modules.cpp
```diff
@@ -838,6 +839,25 @@ PVOID WINAPI DetourFindPayloadEx(_In_ REFGUID rguid,
     return NULL;
 }
 
+BOOL WINAPI DetourFreePayload(_In_ PVOID pvData)
+{
+	BOOL fSucceeded = FALSE;
+
+	HMODULE hModule = DetourGetContainingModule(pvData);
+	assert(hModule != NULL);
```
sylveon, line 847: The asserts would break a caller's error handling because it possibly changes GetLastError.

src/modules.cpp
```diff
@@ -838,6 +839,25 @@ PVOID WINAPI DetourFindPayloadEx(_In_ REFGUID rguid,
...
+	if (hModule != NULL)
+	{
+		fSucceeded = VirtualFree(hModule, 0, MEM_RELEASE);
+		assert(fSucceeded);
+		if (fSucceeded)
+		{
+			hModule = NULL;
```
sylveon, line 854: Not sure why this is needed - it's a local variable. I would suggest

```cpp
BOOL WINAPI DetourFreePayload(_In_ PVOID pvData)
{
	BOOL fSucceeded = FALSE;

	HMODULE hModule = DetourGetContainingModule(pvData);
	if (hModule != NULL)
	{
		fSucceeded = VirtualFree(hModule, 0, MEM_RELEASE);
	}

	return fSucceeded;
}

File: src/modules.cpp [hunk 3]
```diff
@@ -884,6 +904,11 @@ BOOL WINAPI DetourRestoreAfterWithEx(_In_reads_bytes_(cbData) PVOID pvData,
         }
         VirtualProtect(pder->pidh, pder->cbidh, dwPermIdh, &dwIgnore);
     }
+	//Delete the payload after successful recovery to prevent repeated restore
+	if (fSucceeded)
+	{
+        DetourFreePayload(pder);
```
sylveon, line 910: Tabs for indentation here too.

src/creatwth.cpp
```diff
@@ -11,6 +11,7 @@
 #define DETOURS_INTERNAL
 #include "detours.h"
 #include <stddef.h>
+#include <assert.h>
```
sylveon, line 14: That include isn't needed anymore

src/modules.cpp
sonyps5201314: Reset pointer variable is a good habit for program, and it is not important. It won't affect the rest of the place.

src/creatwth.cpp
sonyps5201314: The latest commit has resolve it.

src/modules.cpp
sonyps5201314: The latest commit has resolved it.

src/modules.cpp
sonyps5201314: The latest commit has resolved it.

src/modules.cpp
sonyps5201314: The latest commit has resolved it.

src/creatwth.cpp
```diff
@@ -1019,6 +1020,12 @@ VOID CALLBACK DetourFinishHelperProcess(_In_ HWND,
         rlpDlls = NULL;
     }
 
+	//Delete the payload after execution to release the memory occupied by it
+	if (s_pHelper != NULL)
+	{
+        DetourFreePayload(s_pHelper);
```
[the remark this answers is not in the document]
sonyps5201314: The latest commit has resolved it.

src/detours.h
```diff
@@ -47,6 +47,7 @@
 #include <strsafe.h>
 #pragma warning(pop)
 #endif
+#include <crtdbg.h>
```
sylveon, line 50: Guard this include under #ifdef _DEBUG?

sonyps5201314: No, crtdbg.h can check _DEBUG marco in its internal,
image
this macro and function is ported from my personal project many year ago,  so it has be tested many times and many years.

image

image

`_CrtDbgReport` is basic of CRT's _ASSERTE, ATL's ATLASSERT, and MFC's ASSERT, but _CrtDbgReportW can only be avaliable in high version visual c, so I decide to use _CrtDbgReport at that time.

src/modules.cpp
bgianfo: I still see this on your most recent commit.

sonyps5201314: @bgianfo, Do you mean "Tabs for indentation"? I checked the code again, but still couldn't find the place you mentioned. Can you point it out for me?
It may be possible to add a [.editorconfig ](https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2019) file to the project to help contributors reduce such formatting requirements

src/creatwth.cpp
```diff
@@ -1019,6 +1019,11 @@ VOID CALLBACK DetourFinishHelperProcess(_In_ HWND,
         rlpDlls = NULL;
     }
 
+    // Delete the payload after execution to release the memory occupied by it
+    if (s_pHelper != NULL) {
+        DetourFreePayload(s_pHelper);
```
bgianfo, line 1024: We should set s_pHelper to NULL here?

src/modules.cpp
bgianfo: @sonyps5201314 sure I can investigate adding that.

src/creatwth.cpp
```diff
@@ -1019,6 +1019,11 @@ VOID CALLBACK DetourFinishHelperProcess(_In_ HWND,
         rlpDlls = NULL;
     }
 
+    // Delete the payload after execution to release the memory occupied by it
```
bgianfo, line 1022: I think it would be useful to clarify this comment a bit, mentioning where `s_pHelper` comes from?
Maybe something like: 

Note: s_pHelper is allocated as part of injecting the payload in `DetourCopyPayloadToProcess(..)`, it's a fake section and not data allocated by the loader.

src/modules.cpp
```diff
@@ -838,6 +838,23 @@ PVOID WINAPI DetourFindPayloadEx(_In_ REFGUID rguid,
     return NULL;
 }
 
+BOOL WINAPI DetourFreePayload(_In_ PVOID pvData)
+{
+    BOOL fSucceeded = FALSE;
+
+    HMODULE hModule = DetourGetContainingModule(pvData);
```
bgianfo, line 845: I think this deserves a comment describing that you are (cleverly) using `DetourGetContainingModule(..)` to get the root of the `VirtualAlloc` that was created in `DetourCopyPayloadToProcess(..)`. Otherwise it's going to take the reader a bit to figure out what's going on here, and potentially looking like a bug.

src/creatwth.cpp
bgianfo: Maybe we should also add a comment IN `DetourCopyPayloadToProcess(..)` noting that the `VirtualAlloc` is free'd with `DetourFreePayload(..)`.

bgianfo (changes requested):
Looks good, I think it's ready for merge once these minor things are addressed. Sorry for the wait!

sylveon:
Since this PR adds special behavior with the _DEBUG macro, we might want to have Debug build configs set the `_DEBUG` macro now

src/modules.cpp
sonyps5201314: Good idea, the latest submission has been changed in accordance with your suggestions.

src/creatwth.cpp
sonyps5201314: Good idea, the latest submission has been changed in accordance with your suggestions.

src/creatwth.cpp
sonyps5201314: Good idea, the latest submission has been changed in accordance with your suggestions.

bgianfo (approved):

PR #144, opened 2020-09-04
fix if the hook is not called at the start of the process, the hook may fail

sonyps5201314:
for example, create a c++ console project, create many threads to call CreateFileW,ReadFile,CloseHandle
and create serveral threads to execute Hook and Unhook CreateFileW and ReadFile function every 100 milliseconds. 
it will be failed after some time.


bgianfo:
Thanks for the contribution! 

It looks like you might have mistakenly included all of the other commits from your changes here?

bgianfo (changes requested):
Thanks for the contribution!

It looks like you might have mistakenly included all of the other commits from your changes here?

sonyps5201314:

bgianfo (changes requested):

src/detours.cpp
```diff
@@ -1515,6 +1548,8 @@ struct DetourOperation
     ULONG               dwPerm;
 };
 
+static HANDLE				s_hHeap = NULL;
```
bgianfo, line 1551: Never used?

src/creatwth.cpp
```diff
@@ -1080,7 +1080,7 @@ BOOL WINAPI AllocExeHelper(_Out_ PDETOUR_EXE_HELPER *pHelper,
         cSize += (DWORD)cchDest + 1;
     }
 
-    Helper = (PDETOUR_EXE_HELPER) new NOTHROW BYTE[sizeof(DETOUR_EXE_HELPER) + cSize];
+    Helper = (PDETOUR_EXE_HELPER)DetourCreateObjectArray<BYTE>(sizeof(DETOUR_EXE_HELPER) + cSize);
```
bgianfo, line 1083: Why can't we just pass `PDETOUR_EXE_HELPER` to as the template parameter? 
```suggestion
    Helper = (PDETOUR_EXE_HELPER)DetourCreateObjectArray<PDETOUR_EXE_HELPER>(sizeof(DETOUR_EXE_HELPER) + cSize);
```

src/detours.cpp
```diff
@@ -18,6 +18,22 @@
 
 #define NOTHROW
 
+//////////////////////////////////////////////////////////////////////////////
+//
+
+extern "C" void __cdecl _assert(const char* expr, const char* filename, unsigned lineno);
+void DetourAssert(const char* expr, const char* filename, unsigned lineno)
```
bgianfo, line 25: You introduced a seperate Assert method in one of your other PRs, can you figure out which one you want to contribute and stanardize on that?

src/detours.cpp
```diff
@@ -1959,12 +2013,287 @@ LONG WINAPI DetourUpdateThread(_In_ HANDLE hThread)
...
+    // get a view of the threads in the system
+
+    if (!CreateProcessSnapshot(snapshot)) {
+        DETOUR_TRACE(("detours: can't get process snapshot!"));
+        return FALSE;
+    }
+
+    // this never returns NULL as the current process id is always present in the processes snapshot
+    *procInfo = FindProcess(*snapshot, GetCurrentProcessId());
+    DETOUR_ASSERT(procInfo);
```
bgianfo, line 2244: This ass asserting the wrong pointer value...

src/detours.cpp
```diff
@@ -1959,12 +2013,287 @@ LONG WINAPI DetourUpdateThread(_In_ HANDLE hThread)
...
+typedef LONG NTSTATUS;
+
+typedef LONG KPRIORITY;
+
+typedef enum _SYSTEM_INFORMATION_CLASS
+{
+    SystemBasicInformation = 0,
+    SystemPerformanceInformation = 2,
+    SystemTimeOfDayInformation = 3,
+    SystemProcessInformation = 5,
```
bgianfo, line 2036: Only declare what we need?

src/detours.h
```diff
@@ -539,14 +541,26 @@ typedef BOOL (CALLBACK *PF_DETOUR_IMPORT_FUNC_CALLBACK_EX)(_In_opt_ PVOID pConte
 typedef VOID * PDETOUR_BINARY;
 typedef VOID * PDETOUR_LOADED_BINARY;
 
+//////////////////////////////////////////////////////////// Memory management APIs.
+BOOL WINAPI DetourCreateHeap(BOOL fAutoDestroy);
+HANDLE WINAPI DetourGetHeap();
```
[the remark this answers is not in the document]
sonyps5201314: Yes, it's OK. This modification has expired, and the API is no longer included in the latest commit.

src/detours.cpp
```diff
@@ -1959,12 +2040,303 @@ LONG WINAPI DetourUpdateThread(_In_ HANDLE hThread)
...
 
     t->hThread = hThread;
+	t->fCloseThreadHandleOnDestroyDetourThreadObject = fCloseThreadHandleOnDestroyDetourThreadObject;
     t->pNext = s_pPendingThreads;
     s_pPendingThreads = t;
 
     return NO_ERROR;
 }
 
+//Code modified from https://github.com/apriorit/mhook/blob/master/mhook-lib/mhook.c
```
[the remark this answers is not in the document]
sonyps5201314: The source project is also released under the MIT license, so it should not conflict with this project. If you are still worried, I can write the same function code again in my own way.

src/detours.h
```diff
@@ -539,14 +541,26 @@ typedef BOOL (CALLBACK *PF_DETOUR_IMPORT_FUNC_CALLBACK_EX)(_In_opt_ PVOID pConte
...
+void WINAPI DetourDestroyHeap();
+#ifdef __AUTO_CREATE_DETOUR_HEAP__
+static BOOL s_fAlreadyCreatedDetourHeap = DetourCreateHeap(TRUE);
```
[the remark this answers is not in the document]
sonyps5201314: The code here has expired, and this API is no longer needed in my latest optimized commit.

src/detours.cpp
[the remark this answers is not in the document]
sylveon: The MIT license requires complete attribution, by also including the copyright noticec and a copy of the license file. So I can understand why Detours would want to avoid that.

src/detours.h
```diff
@@ -1118,6 +1132,97 @@ BOOL WINAPI DetourVirtualProtectSameExecute(_In_  PVOID pAddress,
...
+    }
+
+    ::new(p) T(p1);
+
+    return p;
+}
+template<class T, class P1, class P2>
+T* DetourCreateObject(P1 p1, P2 p2)
+{
+    T* p = (T*)HeapAlloc(DetourGetHeap(), 0, sizeof(T));
```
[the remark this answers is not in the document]
sonyps5201314: But not all the codes in the Detours library are in C++ classes. Many of them are organized as functions. I don’t know how to overload operator new/delete globally, and Detours can be compiled into a static library. Will the overloaded new/delete global operator affect other user codes?

src/detours.cpp
```diff
@@ -1524,6 +1543,43 @@ static PVOID *              s_ppPendingError        = NULL;
...
+
+BOOL WINAPI DetourCreateHeap(BOOL fAutoDestroy)
+{
+    BOOL bResult = FALSE;
+    if (s_hHeap == NULL) {
+        s_hHeap = HeapCreate(0, 0, 0);
+        assert(s_hHeap);
+        if (s_hHeap != NULL) {
+            if (fAutoDestroy) {
+                atexit(DetourDestroyHeap__for__atexit);
```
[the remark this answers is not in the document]
sonyps5201314: Yes, these codes here has expired, and these code are no longer existed in my latest optimized commit.

src/detours.cpp
sonyps5201314: Yes, these codes here has expired, and these code are no longer existed in my latest optimized commit.

src/creatwth.cpp
sonyps5201314: Because DETOUR_EXE_HELPER is only the size of the header of the payload, plus cSize*sizeof(BYTE) is the entire size of the payload. If write it as you said, it just means that the memory space of the size of cSize*sizeof(PDETOUR_EXE_HELPER) is allocated. PDETOUR_EXE_HELPER is just a pointer type, so its size is only 4 or 8. So this is not in line with the original intention.

src/detours.cpp
sonyps5201314: These are also declared in `"C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\um\winternl.h"`, so you should not worry about leaking too much information in the Windows kernel to users. These are kept just to be consistent with the original source of the code, of course, I can change it if you feel unnecessary.

src/detours.cpp
sonyps5201314: Yes, this looks like a typo in [the original code](https://github.com/apriorit/mhook/commit/29ef9743a94cc811de29f4d65d37f3769c5cec2d).

src/detours.cpp
sonyps5201314: This one, the one in the other PR can only be used in the _DEBUG environment, and this one is unlimited. Can also be used in a publishing environment. It's just that the commit was earlier, it should be covered by this, guess there should be no conflict, if there is a merge conflict, you can tell me, and I will change it.

src/detours.cpp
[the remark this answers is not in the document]
sonyps5201314: OK, I have refine the code to remove use the code from mhook.

src/detours.cpp
sonyps5201314: the latest commit have remove the code from mhook.

src/detours.cpp
sonyps5201314: the latest commit have remove the code from mhook.

src/creatwth.cpp
bgianfo: Sure, read the code wrong. Thanks!

src/detours.cpp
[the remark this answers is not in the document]
bgianfo: thanks

bgianfo:
Hey @sonyps5201314 thanks again for the contribution.

@dtarditi and I looked at this togeather today, and we decided the change as it is, is very risk to merge since it's so large and invasive. Can you please desribe what the actual race condition / bug this is fixing is so we can work on trying to scope down this change? It's not clear from the description or the commits where the bug you are fixing actually.

Is it also possible to introduce a stress test / unit test to validate we have fix the issue and don't regress it in the future?

sonyps5201314:
[TestDetoursBug144.zip](https://github.com/microsoft/Detours/files/6095398/TestDetoursBug144.zip) is a BUG reproduction program.
Compile, debug and run the `DebugMaster` configuration, the program will crash immediately, because it uses the official unfixed `master` branch code.
image

Compile, debug and run the `DebugMyFixedBranch` configuration, the program will run normally, because it uses the fixed `fix_maybe_can_not_hook_functions_not_at_the_starttime_of_a_process code` in my PR.

Note: It is recommended to use VS2013 to compile and debug the program of this project, because VS2019 and VS2017 have the bug that the call stack of the program cannot be viewed when the x86 program crashes or deadlocks. It has been reported for almost a year, and the official hasn't repaired it yet.
[vs2017 and vs2019 can`t show full call stack when a deadlock occurs in x86 c++.](https://developercommunity.visualstudio.com/t/vs2017-and-vs2019-cant-show-full-call-s/1182801)

fredemmott:
> @bgianfo Can you please desribe what the actual race condition / bug this is fixing

The original commit seems to mostly be related to #70 (though covers a broader problem): any new/delete/malloc/free etc after any `DetourUpdateThread()` is risky:

In windows, there is a per-heap lock, which new/delete/malloc/free always need to require, and in most applications, practically all operations will be on the same default heap - in effect, roughly speaking, there's a per-process lock on heap operations. See https://docs.microsoft.com/en-us/windows/win32/memory/heap-functions

`DetourUpdateThread()` ultimately calls `SuspendThread()`; after this point, if the thread you've passed in happens to already have the heap lock, nothing else can acquire the lock, and all new/delete/malloc/free will deadlock. This includes:
* the explicit `new` operations in `DetourUpdateThread()` - e.g. if you're updating two threads, and the first one has the heap lock, the second `DetourUpdateThread()` will deadlock
* the malloc inside the printfs inside `DETOUR_TRACE`

You can reproduce this by:
- creating 3 threads
- in thread 1, call `HeapLock(GetProcessHeap())`
- just loop in thread 2
- in thread 3, call `DetourTransactionBegin(); DetourUpdateThread(thread_1_handle); DetourUpdateThread(thread_2_handle);`

--

There are also several other bugs fixed in later commits; I'm not came across them in practice, so I'm not able to explain them.

--

As for how this relates to 'at the start of the process':

- updating all threads is a relatively common (and necessary) pattern, which significantly raises the risk of hitting this condition. This PR adds a `DetourUpdateAllOtherThreads()` function to make this pattern easier, but that's tangental to the bug fix.

- if you attach on process startup, there are *many* more mallocs/news happening than later in the process lifecycle: as well as the program code, there's all the implicit ones behind `LoadLibrary()` etc. Again, this significantly raises the risk

fredemmott:
Here is a small standalone example: https://gist.github.com/fredemmott/adb9c41d7cff44aecf92bece990f4c3f

fredemmott:
An alternative fix for that first issue would be to simply call `HeapLock(GetProcessHeap())` in `DetourTransactionBegin()`, and `HeapUnlock(GetProcessHeap())` in `DetourTransactionCommit()` + `DetourTransactionAbort()`

sonyps5201314:
The latest committed code makes Detours use VirtualAlloc/VirtualFree for memory allocation and deallocation, so there is no Heap lock problem you worry about.
```cpp
// After calling DetourUpdateAllOtherThreads, you should call DetourTransactionCommit(Ex) or DetourTransactionAbort as soon as possible.
// In addition to the DetourAttach(Ex)\DetourDetach(Ex) call, other user operations should not be included between them,
// because other user operations may cause CRT lock competition, and at this time CRT lock may be owned by other threads.
// Other threads have been suspended at this time, so that may cause deadlock problems
LONG WINAPI DetourUpdateThreadEx(_In_ HANDLE hThread, _In_opt_ BOOL fCloseThreadHandleOnDestroyDetourThreadObject /*= TRUE*/);
BOOL WINAPI DetourUpdateAllOtherThreads();
```
and you can read the commented code above for DetourUpdateThreadEx and DetourUpdateAllOtherThreads to understand the precautions for using them. and from comments you can know that `HeapLock(GetProcessHeap()) ...` is not safe enough.
and you must know that the purpose of DetourUpdateAllOtherThreads is to solve the problem that the HOOKed instruction is being run by other threads, not the Heap lock problem. Heap and CRT lock problems are only one problem that may be caused by DetourUpdateAllOtherThreads. Therefore, I introduced DetourMemAlloc, DetourMemReAlloc, DetourMemFree to avoid this problem. The code in the PR has also been used in many programs of my own and my company and has been running stably in many computers for a long time, so you can believe that they are reliable of.

fredemmott:
> But if a thread already owns the lock and has been suspended by other threads of the user, you call HeapLock((HANDLE)_get_heap_handle()); it will only make the current thread fall into an infinite wait or deadlock

This can generally be avoided by calling `HeapLock()` before suspending *any* threads.If something else in the process has already locked a thread which holds the lock, either that thread will eventually be resumed (in which case it's not an infinite wait), or the app will be hanging forever anyway. Either way, even if we can install the hook while something else has the heap lock, my detours (and the application code that calls them) still use the heap, so it isn't a practical concern, as long as my thread acquires a heap lock before suspending *any* threads.

> so The best way is to avoid using the HeapAlloc family of functions for memory allocation.

I don't disagree on this, but it is not necessary for many cases.

> The code in the PR has also been used in many programs of my own and my company and has been running stably in many computers for a long time, so you can believe that they are reliable of.

I'm looking for minimal - and understandable - changes on top of master. This PR fixes several issues (and I understand this is more than the heaplock race), and adds new features. Regardless of reliability for you, that makes it not fit my goals. Splitting it up with clear explanations would also likely aid Microsoft in reviewing and ideally merging.

sonyps5201314:
```cpp
DWORD WINAPI ThreadProc_UserCode(LPVOID lpThreadParameter)
{
	Sleep(1000);
	HeapLock((HANDLE)_get_heap_handle());
	Sleep(10000000);
	HeapUnlock((HANDLE)_get_heap_handle());
	return 0;
}

DWORD WINAPI ThreadProc_TestHook(LPVOID lpThreadParameter)
{
	Sleep(2000);
	while (1)
	{
		OutputDebugStringA("ThreadProc_TestHook prepare for hook.\n");
		//replace "OutputDebugStringA" to "printf" function will trigger another deadlock

		HeapLock((HANDLE)_get_heap_handle());

		OutputDebugStringA("ThreadProc_TestHook start to hook.\n");
		DetourTransactionBegin();

		DetourUpdateThread(GetCurrentThread());

		DetourTransactionCommit();

		HeapUnlock((HANDLE)_get_heap_handle());
	}
	return 0;
}
int main()
{
	HANDLE hThread = CreateThread(NULL, 0, ThreadProc_UserCode, NULL, 0, NULL);
	if (hThread)
	{
		CloseHandle(hThread);
		hThread = NULL;
	}
	Sleep(100);

	hThread = CreateThread(NULL, 0, ThreadProc_TestHook, NULL, 0, NULL);
	if (hThread)
	{
		CloseHandle(hThread);
		hThread = NULL;
	}
	Sleep(100 * 1000);
}
```
The above is a simple example that doesn't do what you expect.
In addition to our own controllable code, a program may also contain system library code, and code injected by third-party DLLs. The code in `ThreadProc_UserCode` may appear in third-party DLLs and system libraries. For example, in AcLayers.dll, it hooks the Heap series functions and has its own lock implementation for the Heap functions, so you can't think that the call to HeapLock returns, you won't Stuck in the next calls to functions such as HeapAlloc.

> I don't disagree on this, but it is not necessary for many cases.

Detours, as a general-purpose library program, must be able to accommodate all situations, not just most of them.

Issue #116, opened 2020-08-06
[syelogd] Negative timing difference between successive events with option /d (delta time) 


neimad29b1:
It may happen that timing difference between two successive events (issued from different threads) is negative.
The problem is actually that with option /d, the delta shown is incorrect (**1844674** instead of **-x**).

`      1 4372 50.60: trc: 010 ..........................`
`1844674 4372 50.60: trc: 018 ..........................`
`      2 4372 50.60: trc: 010 ..........................`

This can be simply corrected using a signed integer at line 96 of **syelogd.cpp**
`LONG64 delta = ul.QuadPart - s_llLastTime;`

bgianfo:
Nice catch @neimad29b1 tagging this in case someone wants to pick this up before we get a chance.

warlockz:
Please can you review for any comments or merge my PR

PR #146, opened 2020-09-29
Fixed#116 via changing to signed for negative diff

warlockz:
Addressed : Issue #116 [syelogd] Negative timing difference between successive events with option /d (delta time).

Please can you review and merge the fix.

Thanks 👍


bgianfo:
Thanks @warlockz

PR #173, opened 2021-02-16
Prevent conflict with MINGW32 defines

ReneNyffenegger:
mingw defines (typedefs?) `ULONG_PTR` and `LONG_PTR`, but not `_MSC_VER`. So the `typedef LONG LONG_PTR` in `detours.h` gives an error when compiling with mingw.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/173)

### DIFF
--- a/src/detours.h
+++ b/src/detours.h
@@ -126,7 +126,7 @@
 //////////////////////////////////////////////////////////////////////////////
 //
 
-#if (_MSC_VER < 1299)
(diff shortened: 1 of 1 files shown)


bgianfo:
Thanks!

PR #182, opened 2021-03-05
Fix potential integer overflow

dtarditi:
When injecting a DLL into a process, it is possible that the process memory  has been corrupted.  The values in the import table for the process could be incorrect, which could cause an integer overflow when calculating the size of the new import table.  Add code to protect against this to UPDATE_IMPORTS_XX.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/182)

### DIFF
--- a/src/detours.h
+++ b/src/detours.h
@@ -36,6 +36,7 @@
 #pragma warning(disable:6102 6103) // /analyze warnings
 #endif
 #include <strsafe.h>
+#include <intsafe.h>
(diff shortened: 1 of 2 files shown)


bgianfo (approved):
:shipit:

AraHaan:
o my so this been a thing for years?

PR #175, opened 2021-02-27
Enable debug logging on 'Debug' build configs

sylveon:
This enables 'Debug' build configs to build the debug output and breakpoints, and fixes all mistakes that slipped in printf format specifiers.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/175)

PR #184, opened 2021-03-06
Fix: Address print strings regression from PR #175

### LINKED ISSUE #183: Confusingly large number of changes
In [PR175](https://github.com/microsoft/Detours/pull/175), In many `DETOUR_TRACE `calls related to `GetLastError()`, the formatting parameter `%d` is replaced with `%lx`, but from the `WinError.h` file of the Windows SDK, we can know that these error values ​​are all decimal numbers, so using `%u` will make the code It's easier to understand, and `%lx` hexadecimal will confuse users.
original code:
```c++
        DETOUR_TRACE(("... failed: %d\n", GetLastError()));
```
error code in Windows SDK:
```c++
#define ERROR_INVALID_PARAMETER          87L    // dderror
```

sylveon:
Fixes #183

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/184)

### DIFF
--- a/src/creatwth.cpp
+++ b/src/creatwth.cpp
@@ -52,7 +52,7 @@ static PVOID LoadNtHeaderFromProcess(_In_ HANDLE hProcess,
 
     IMAGE_DOS_HEADER idh;
     if (!ReadProcessMemory(hProcess, pbModule, &idh, sizeof(idh), NULL)) {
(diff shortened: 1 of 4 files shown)


bgianfo:
Thanks @sylveon

PR #191, opened 2021-03-31
Reduce the executable code size.

jdp1024:
In the original version of Detours, s_rceCopyTable and s_rceCopyTable0F are two big arrays, both consisting of 256 COPYENTRY's, occupying 256*(8+8)*2=8192 bytes. To reduce the executable code size, those two arrays are changed to use the index to a COPYENTRY array instead of the actual entity.

s_rceCopyTable and s_rceCopyTable0F now are arrays of 256 BYTEs, the COPYENTRY array consists of only the 45 unique operands. All these occupy 256*2+45*(8+8)=1232 bytes, thus we save 7060 bytes with almost no performance sacrifice.

### DIFF
--- a/src/disasm.cpp
+++ b/src/disasm.cpp
@@ -195,63 +195,56 @@ class CDetourDis
         COPYFUNC    pfCopy;                 // Function pointer.
     };
 
+    enum {
+        eENTRY_CopyBytes1 = 0,
+        eENTRY_CopyBytes1Address,
+        eENTRY_CopyBytes1Dynamic,
+        eENTRY_CopyBytes2,
+        eENTRY_CopyBytes2Jump,
+        eENTRY_CopyBytes2CantJump,
+        eENTRY_CopyBytes2Dynamic,
+        eENTRY_CopyBytes3,
+        eENTRY_CopyBytes3Dynamic,
+        eENTRY_CopyBytes3Or5,
+        eENTRY_CopyBytes3Or5Dynamic,
+        eENTRY_CopyBytes3Or5Rax,
+        eENTRY_CopyBytes3Or5Target,
+        eENTRY_CopyBytes4,
+        eENTRY_CopyBytes5,
+        eENTRY_CopyBytes5Or7Dynamic,
+        eENTRY_CopyBytes7,
+        eENTRY_CopyBytes2Mod,
+        eENTRY_CopyBytes2ModDynamic,
+        eENTRY_CopyBytes2Mod1,
+        eENTRY_CopyBytes2ModOperand,
+        eENTRY_CopyBytes3Mod,
+        eENTRY_CopyBytes3Mod1,
+        eENTRY_CopyBytesPrefix,
+        eENTRY_CopyBytesSegment,
+        eENTRY_CopyBytesRax,
+        eENTRY_CopyF2,
+        eENTRY_CopyF3,
+        eENTRY_Copy0F,
+        eENTRY_Copy0F78,
+        eENTRY_Copy0F00,
+        eENTRY_Copy0FB8,
+        eENTRY_Copy66,
+        eENTRY_Copy67,
+        eENTRY_CopyF6,
+        eENTRY_CopyF7,
+        eENTRY_CopyFF,
+        eENTRY_CopyVex2,
+        eENTRY_CopyVex3,
+        eENTRY_CopyEvex,
+        eENTRY_CopyXop,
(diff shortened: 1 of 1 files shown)


bgianfo:
@jaykrell I think you were working on this code in the recent past? Do you have any thoughts on this?

jaykrell:
It is a micro optimization but yes makes sense just from the description.
There are 512 multi-word entries, composed of only a few unique values.

src/disasm.cpp
```diff
@@ -586,24 +580,24 @@ PBYTE CDetourDis::Invalid(REFCOPYENTRY pEntry, PBYTE pbDst, PBYTE pbSrc)
...
     return (this->*pEntry->pfCopy)(pEntry, pbDst + 1, pbSrc + 1);
 }
 
 PBYTE CDetourDis::Copy0F78(REFCOPYENTRY, PBYTE pbDst, PBYTE pbSrc)
 {
     // vmread, 66/extrq, F2/insertq
 
-    static const COPYENTRY vmread = /* 78 */ ENTRY_CopyBytes2Mod;
-    static const COPYENTRY extrq_insertq = /* 78 */ ENTRY_CopyBytes4;
+    const BYTE vmread = /* 78 */ eENTRY_CopyBytes2Mod;
```
jaykrell, line 591: enum instead of byte?

src/disasm.cpp
```diff
@@ -660,7 +654,7 @@ PBYTE CDetourDis::CopyF6(REFCOPYENTRY pEntry, PBYTE pbDst, PBYTE pbSrc)
 
     // TEST BYTE /0
     if (0x00 == (0x38 & pbSrc[1])) {    // reg(bits 543) of ModR/M == 0
-        static const COPYENTRY ce = /* f6 */ ENTRY_CopyBytes2Mod1;
+        const COPYENTRY ce = /* f6 */ s_rceCopyMap[eENTRY_CopyBytes2Mod1];
```
jaykrell, line 657: reference or pointer instead of value?

src/disasm.cpp
```diff
@@ -953,543 +1007,543 @@ const BYTE CDetourDis::s_rbModRm[256] = {
...
-const CDetourDis::COPYENTRY CDetourDis::s_rceCopyTable[] =
+const BYTE CDetourDis::s_rceCopyTable[] =
 {
-    /* 00 */ ENTRY_CopyBytes2Mod,                      // ADD /r
-    /* 01 */ ENTRY_CopyBytes2Mod,                      // ADD /r
-    /* 02 */ ENTRY_CopyBytes2Mod,                      // ADD /r
-    /* 03 */ ENTRY_CopyBytes2Mod,                      // ADD /r
-    /* 04 */ ENTRY_CopyBytes2,                         // ADD ib
-    /* 05 */ ENTRY_CopyBytes3Or5,                      // ADD iw
+    /* 00 */ eENTRY_CopyBytes2Mod,                      // ADD /r
```
jaykrell, line 1012: You might rephrase the PR with the same intent and even resulting bytes, but without changing these lines. i.e. just to make the diff easier to read.

jaykrell: Alternatively, someone else might try to make the same change and verify they get the same result. I understand, you just prepended 'e' to every identifer, over 512 lines, it is just hard to verify.

jaykrell:
I should point out. I changed all of this code so the main handler function is templatized, and the table is filled with (mostly?) instantiated pointers to it. So that ends up around 512*8=2k and perhaps faster, perhaps a little larger.
That version shipped in a product. Those could also be compressed in your way, to the 512 bytes plus a little extra, and then even smaller tables than yours, but larger code.
Galen's response at the time though, was that Detour's users do not understand C++ templates, so do not use them.

src/disasm.cpp
jaykrell: Eh, byte is maybe best and certainly works.

src/disasm.cpp
jdp1024: This is what `diff -ur` or `git diff` gave, although less straight. Please let me know how to make this more human readable.

I used a regex to generate the eENTRY_* definitions (value/name pair)  from the original ones.

src/disasm.cpp
jdp1024: Of course, I will make another commit.

AtariDreams:
Any updates on this?

PR #209, opened 2021-07-28
Disx86: Add support for FP16.

jaykrell:
See https://software.intel.com/content/www/us/en/develop/download/intel-avx512-fp16-architecture-specification.html

For 16bit floating point numbers, EVEX maps 5 and 6 become defined.
EVEX includes a 4 bit map.
Previously only 2 bits were allowed. Now 3.

EVEX and VEX share logic but VEX cannot encode FP16.

Tests are lacking.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/209)


jaykrell:
@bgianfo 
I am not able to add reviewers.

bgianfo:
Is there a way we can add unit tests for the disassembler?

jaykrell:
> Is there a way we can add unit tests for the disassembler?

Extensive testing was present in the 3.0 tree.
It disassembled a bunch of code with msdis and detours and compared.
It found bugs I think in both, albeit maybe only very obscure in msdis, like the x86-to-ia64 call instruction.
I modified the tests to be runnable cross platform.

PR #211, opened 2021-08-17
DetourAttachEx: set out params to NULL on failure

matbech:
Sets ppRealDetour and ppRealTarget out params to NULL on failure.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/211)


bgianfo:
Thanks! 👍

PR #225, opened 2022-03-02
Enable compilation time configuration of maximum section count and expose DetourIsFunctionImported

JohnMcPMS:
A few minor changes:
1. Expose `detour_is_imported` via a new public function `DetourIsFunctionImported`
2. Make certain runtime checks happy by masking `pbTarget` before casting to smaller type
3. Enable user to compile with a different number of supported section headers (leaves default at 32)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/225)

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -1395,6 +1395,12 @@ PVOID WINAPI DetourAllocateRegionWithinJumpBounds(_In_ LPCVOID pbTarget,
     return pbNewlyAllocated;
 }
 
+BOOL WINAPI DetourIsFunctionImported(_In_ PBYTE pbCode,
+                                     _In_ PBYTE pbAddress)
+{
+    return detour_is_imported(pbCode, pbAddress);
+}
+
 static PDETOUR_TRAMPOLINE detour_alloc_trampoline(PBYTE pbTarget)
 {
     // We have to place trampolines within +/- 2GB of target.
@@ -1437,7 +1443,8 @@ static PDETOUR_TRAMPOLINE detour_alloc_trampoline(PBYTE pbTarget)
     // We need to allocate a new region.
 
     // Round pbTarget down to 64KB block.
-    pbTarget = pbTarget - (PtrToUlong(pbTarget) & 0xffff);
+    // /RTCc RuntimeChecks breaks PtrToUlong.
+    pbTarget = pbTarget - (ULONG)((ULONG_PTR)pbTarget & 0xffff);
 
     PVOID pbNewlyAllocated =
         detour_alloc_trampoline_allocate_new(pbTarget, pLo, pHi);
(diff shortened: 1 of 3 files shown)


src/detours.cpp
```diff
@@ -1395,6 +1395,12 @@ PVOID WINAPI DetourAllocateRegionWithinJumpBounds(_In_ LPCVOID pbTarget,
     return pbNewlyAllocated;
 }
 
+BOOL WINAPI DetourIsFunctionImported(_In_ PBYTE pbCode,
```
bgianfo, line 1398: It would be nice if you could add new tests for `DetourIsFunctionImported` as it's now exposing public functionality. I would suggest some unit tests under [tests/](https://github.com/microsoft/Detours/tree/master/tests).

bgianfo (changes requested):
Can you please squash the PR feedback commits into the original change? So we can avoid squashing everything togeather once this is merged?

src/detours.cpp
JohnMcPMS: I've been trying to find something to do this without just pulling a value out of the IAT, but is it possible to write some expression to get the address of the address table entry in C/C++?  My expectation is that it is mostly a hidden implementation detail, but hopefully there is some way of getting it using MSVC specifics.

```
// Generates load from IAT rather than being address of IAT
auto ptr = &SetLastError;

// If only it were this easy
auto IAT_ptr = &&SetLastError;
```

jaykrell:
```cpp
// Get the address of static imports, w/o relying on __declspec(dllimport)
// or LoadLibrary/GetProcAddress. Consider using decltype(&foo) __imp_foo.
#ifdef _X86_
#pragma warning(disable:4483) // disable warning/error about __identifier(<string>)
#define __imp_DbgPrint                          _imp__DbgPrint
#define __imp_LoadLibraryW                   __identifier("_imp__LoadLibraryW@4")
#endif

extern size_t const __imp_DbgPrint;
extern size_t const __imp_LoadLibraryW;
```
etc.

JohnMcPMS:
`DetourEnumerateModules` appears to be flaky; failing in one flavor last build and two this time (both debug/2019).

```
-------------------------------------------------------------------------------
DetourEnumerateModules
  Passing nullptr, results in current module being returned
-------------------------------------------------------------------------------
test_module_api.cpp(237)
...............................................................................

test_module_api.cpp(243): PASSED:
  REQUIRE( GetLastError() == 0L )
with expansion:
  0 == 0

test_module_api.cpp(244): FAILED:
  REQUIRE( mod == reinterpret_cast<HMODULE>(&__ImageBase) )
with expansion:
  0x005b0000 == 0x01260000

0.000 s: Passing nullptr, results in current module being returned
0.000 s: DetourEnumerateModules
-------------------------------------------------------------------------------
```

None of my changes should have affected this function.  Maybe ASLR is enabled on the 2019 server images?  This test expects that the first module in memory is the exe, and given the behavior it seems the most likely culprit.

jaykrell:
That sounds plausible.
The base address of .exes used to be quite predictable.

bgianfo:
I sent out PR #227 to fix the test, thanks for pointing that out!

JohnMcPMS:
@bgianfo , what changes are still needed?

bgianfo (approved):

bgianfo:
> @bgianfo , what changes are still needed?

Sorry, no re-review was requested, so I didn't know it was ready. 
Merged, thanks!

PR #232, opened 2022-04-10
Fix: Eliminate possible deadlock related to memory allocation on global heap

adams85:
Addresses https://github.com/microsoft/Detours/issues/70 by utilizing the `HeapLock`/`HeapUnlock` APIs as discussed in the issue.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/232)

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -1601,6 +1601,12 @@ _Benign_race_end_
     s_pPendingThreads = NULL;
     s_ppPendingError = NULL;
 
+    if (!HeapLock(GetProcessHeap())) {
+        s_nPendingError = GetLastError();
+
+        return s_nPendingError;
+    }
+
     // Make sure the trampoline pages are writable.
     s_nPendingError = detour_writable_trampoline_regions();
 
@@ -1645,10 +1651,19 @@ LONG WINAPI DetourTransactionAbort()
         delete t;
         t = n;
     }
+
+    // s_nPendingError should not be overwritten as DetourTransactionCommitEx relies on it.
+    LONG nReportedError = NO_ERROR;
+
+    if (!HeapUnlock(GetProcessHeap())) {
+        // This may mask a preceding error but we can't do anything about it as we can report a single error.
+        s_nPendingError = nReportedError = GetLastError();
+    }
+
     s_pPendingThreads = NULL;
     s_nPendingThreadId = 0;
 
-    return NO_ERROR;
+    return nReportedError;
 }
 
 LONG WINAPI DetourTransactionCommit()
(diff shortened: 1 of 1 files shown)


sylveon:
This doesn't solve the issue. Calling HeapLock while a suspended thread has acquired a heap lock by itself will deadlock. The actual resolution would be to create our own heap (using HeapCreate) and allocate everything in that, as this isn't subject to concurrency issue with any suspended thread.

sylveon:
Ah nevermind, I didn't notice this was in DetourTransactionBegin, before any thread could be suspended. This is good then.

src/detours.cpp
```diff
@@ -1648,6 +1654,9 @@ LONG WINAPI DetourTransactionAbort()
     s_pPendingThreads = NULL;
     s_nPendingThreadId = 0;
 
+    // There is nothing we can do if this fails.
+    HeapUnlock(GetProcessHeap());
```
sylveon, line 1658: We should really report this if it fails, because "nobody can use the global heap anymore" seems like a pretty big problem

sylveon: Also, since the heap is locked after s_nPendingThreadId is set in DetourTransactionBegin, we should probably unlock the heap before allowing another thread to start a transaction, so before resetting `s_nPendingThreadId ` here

src/detours.cpp
```diff
@@ -1942,6 +1951,9 @@ typedef ULONG_PTR DETOURS_EIP_TYPE;
         *pppFailedPointer = s_ppPendingError;
     }
 
+    // There is nothing we can do if this fails.
+    HeapUnlock(GetProcessHeap());
```
sylveon, line 1955: Same here

adams85:
Thanks for the feedback. I've updated the PR.

sonyps5201314:
`HeapLock/HeapUnlock` is not a perfect solution, check out the related debate in [this PR](https://github.com/microsoft/Detours/pull/144).

adams85:
So, as I understand, the main problem with the `HeapLock` approach is that there is a race condition: after we have locked the heap in `DetourTransactionBegin`, there could be another thread which may unlock it before we have a chance to suspend that thread. This seems to be a bummer indeed...

It looks like then there's no other foolproof solution than using `HeapCreate` and custom allocation. So I'm closing this in favor of your PR. (I suggest though you clean it up a bit by squashing commits and split it into two parts: one that addresses the deadlock issue and one that introduces `DetourUpdateAllOtherThreads`. So it may have a better chance to get merged.)

sonyps5201314:
@[adams85](https://github.com/adams85)
Thank you for your actions and suggestions, because the title of [my PR](https://github.com/microsoft/Detours/pull/144) is "`Fix if the hook is not called at the start of the process, the hook may fail`", and `deadlock` is part of the issue covered in this issue and `DetourUpdateAllOtherThreads` is the solution part of the way, so I don't think they are easily separable.

PR #233, opened 2022-04-10
Proposal: Introduce DetourUpdateAllOtherThreads

adams85:
Introduce a new public function which updates all threads apart from the current one in a single operation.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/233)

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -1524,6 +1524,28 @@ struct DetourThread
 {
     DetourThread *      pNext;
     HANDLE              hThread;
+    BOOL                fCloseThreadHandleOnDestroy;
+
+    DetourThread()
+    {
+        pNext = NULL;
+        hThread = NULL;
+        fCloseThreadHandleOnDestroy = FALSE;
+    }
+
+    DetourThread(const DetourThread&) = delete;
+    DetourThread& operator= (const DetourThread&) = delete;
+
+    ~DetourThread()
+    {
+        if (hThread) {
+            if (fCloseThreadHandleOnDestroy) {
+                CloseHandle(hThread);
+            }
+
+            hThread = NULL;
+        }
+    }
 };
 
 struct DetourOperation
@@ -1947,6 +1969,11 @@ typedef ULONG_PTR DETOURS_EIP_TYPE;
 
 LONG WINAPI DetourUpdateThread(_In_ HANDLE hThread)
 {
+    return DetourUpdateThreadEx(hThread, FALSE);
+}
+
+LONG WINAPI DetourUpdateThreadEx(_In_ HANDLE hThread, _In_ BOOL fCloseThreadHandleOnDestroy)
+{
     LONG error;
 
     // If any of the pending operations failed, then we don't need to do this.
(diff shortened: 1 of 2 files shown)


src/detours.cpp
```diff
@@ -1524,6 +1524,24 @@ struct DetourThread
 {
     DetourThread *      pNext;
     HANDLE              hThread;
+    BOOL                fCloseThreadHandleOnDestroy;
+
+    DetourThread() {
```
sylveon, line 1529: Allman braces

src/detours.cpp
```diff
@@ -1524,6 +1524,24 @@ struct DetourThread
 {
     DetourThread *      pNext;
     HANDLE              hThread;
+    BOOL                fCloseThreadHandleOnDestroy;
```
sylveon, line 1527: If we're giving this class ownership semantics, delete the copy constructor and assignment operator.

src/detours.cpp
```diff
@@ -1980,12 +2003,57 @@ LONG WINAPI DetourUpdateThread(_In_ HANDLE hThread)
...
+    _In_ HANDLE ThreadHandle,
+    _In_ ACCESS_MASK DesiredAccess,
+    _In_ ULONG HandleAttributes,
+    _In_ ULONG Flags,
+    _Out_ PHANDLE NewThreadHandle
+);
+
+LONG WINAPI DetourUpdateAllOtherThreads()
+{
+    _NtGetNextThread NtGetNextThread = (_NtGetNextThread)GetProcAddress(GetModuleHandle(TEXT("ntdll.dll")), "NtGetNextThread");
```
sylveon, line 2030: Try avoiding undocumented APIs if possible: Windows 8.1 and up have [PssCaptureSnapshot](https://docs.microsoft.com/en-us/windows/win32/api/processsnapshot/nf-processsnapshot-psscapturesnapshot) that you can dynamically query for. Using NtGetNextThread should be fine on older platforms.

src/detours.cpp
```diff
@@ -1980,12 +2003,57 @@ LONG WINAPI DetourUpdateThread(_In_ HANDLE hThread)
...
+        if (!NT_SUCCESS(status)) {
+            if (status != STATUS_NO_MORE_ENTRIES) {
+                DETOUR_TRACE("Failed to enumerate process threads.\r\n");
+                return ERROR_FUNCTION_FAILED;
+            }
+
+            return NO_ERROR;
+        }
+
+        if (currentThreadId != GetThreadId(hThread)) {
```
sylveon, line 2051: GetThreadId was introduced in Vista but Detours supports XP. This won't compile. See #80.

adams85:
I've addressed the first two problems. The other two (regarding the Windows APIs used) would take more time than I can spare now...

sonyps5201314:
I have already introduced the DetourUpdateAllOtherThreads API in [this PR](https://github.com/microsoft/Detours/pull/144) a long time ago. Your code seems to be a version that modified me a long time ago. The NtGetNextThread and GetThreadId you use do not exist in XP, so your code is compatible Question, does not apply to all systems supported by detours. And my code has been repeatedly revised and tested extensively, so it is stable.

sonyps5201314:
@[adams85](https://github.com/adams85)
Thanks very much!

PR #236, opened 2022-04-22
Enable control over import ordinal and add DetourUpdateDllWithDll

JohnMcPMS:
This change is largely about being able to control the import ordinal set in `UPDATE_IMPORTS_XX`, but also includes a new function `DetourUpdateDllWithDll` which is similar to `DetourUpdateProcessWithDllEx`.

Additional parameters for `DetourUpdateProcessWithDllEx` are added to make `DetourUpdateProcessWithDllEx2`.  I wasn't sure what the natural next name was for post-`Ex` (I assume it isn't `ExEx`).  Maybe it should be `DetourUpdateProcessWithDllAndOrdinals` instead?

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/236)


JohnMcPMS:
Needs review.

jaykrell:
Windows has started with "2" and "3", and perhaps, taking parameters in structs.
Though the struct thing..is still wierd -- size or flags, etc.?
Years ago "we" formulated the general case is that you need maybe like: 
flags in
flags honored out
flags ignored out (honored+ignored=in for success)

or perhaps
 flags required in
 flags ignorable in
 flags honored out (of the ignorable ones)

But even this might not suffice -- imagine a caller wants to probe what flags are understood.
Perhaps a simpler and more complete design:
FooGetKnownFlags()
and then caller can decide, and pass no more than that.

Issue #241, opened 2022-06-14
[bugs]ERROR_PIPE_CONNECTED when ConnectNamedPipe in syelogd if there are lots of clients


fishjam:
**Describe the bug**
I have written a UT code with multi thread client to test my pipe server(include the syelog), and I found if there are lots of client request connect to syelog at same time, there is ERROR_PIPE_CONNECTED when call ConnectNamedPipe , ad no err handle for it. 

**Command-line test case**
```
E:\CodeGithub\fishjam-Detours\bin.X64>syelogd.exe
20220614033813631 ---- --.60: Before CreateWorkers.

20220614033813632 ---- --.60: Before CreatePipeConnection

20220614033813632 ---- --.60: Enter  WorkerThread
20220614033813632 ---- --.60: Enter CreatePipeConnection for 220.
20220614033813633 ---- --.60: After CreatePipeConnection

SYELOGD: Ready for clients.  Press Ctrl-C to stop.
20220614033821170 ---- --.60: GetQueuedCompletionStatus, bRet = 1, dwBytes=0, pClient=000001F3E5789390
20220614033821171 ---- --.60: Enter CreatePipeConnection for 220.
20220614033821172 ---- --.00: Error 535 in ConnectNamedPipe.
SYELOGD: Error 535 in ConnectNamedPipe.
```

**Expected behavior**
I fix it by add the speical error handle, and will create the PR for it. 

**Detours version**
*  current master

fishjam:
After fix it:
```
E:\CodeGithub\fishjam-Detours\bin.X64>syelogd.exe
SYELOGD: Ready for clients.  Press Ctrl-C to stop.
20220614052404152 14272 50.60: fishjam test thread:1, testIndex:0
20220614052404152 14272 50.60: fishjam test thread:1, testIndex:1
20220614052404153 14272 50.60: fishjam test thread:1, testIndex:2
20220614052404155 14272 50.60: fishjam test thread:2, testIndex:0
20220614052404155 14272 50.60: fishjam test thread:1, testIndex:3
20220614052404155 14272 50.60: fishjam test thread:2, testIndex:1
20220614052404155 14272 50.60: fishjam test thread:1, testIndex:4
20220614052404155 14272 50.60: fishjam test thread:2, testIndex:2
20220614052404155 14272 50.60: fishjam test thread:1, testIndex:5
20220614052404156 14272 50.60: fishjam test thread:2, testIndex:3
[...]

PR #242, opened 2022-06-14
fix ERROR_PIPE_CONNECTED(535) for ConnectNamedPipe

fishjam:
PR for [ERROR_PIPE_CONNECTED when ConnectNamedPipe in syelogd if there are lots of clients](https://github.com/microsoft/Detours/issues/241)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/242)


samples/syelog/syelogd.cpp
```diff
@@ -180,9 +181,15 @@ PCLIENT CreatePipeConnection(HANDLE hCompletionPort)
         MyErrExit("CreateIoComplietionPort pClient");
     }
 
-    if (!ConnectNamedPipe(hPipe, pClient)) {
-        if (GetLastError() != ERROR_IO_PENDING &&
-            GetLastError() != ERROR_PIPE_LISTENING) {
+    if (!ConnectNamedPipe(hPipe, pClient)) {
+        DWORD dwLastErr = GetLastError();
+        //client already connected while call this fun
```
bgianfo, line 186: Small nitpick, I think this comment could be more clear, something  like: 
```suggestion
        // Handle race between multiple client connections

```

bgianfo (changes requested):
Happy to merge after addressing the minor comment nits. 

Thanks!

bgianfo (changes requested):
It looks like you might have changed the line endings of this file by accident. The entire file appears to be edited. 

Would you mind undo doing that, it makes any diff extremely hard to read.

fishjam:
sorry, I merged your comment change code on web, seems it change the line ending. 
I will try to undo it, if I failed(or mayb I can not find the way), you can change the code and commit. Anyway, as long as we can solve it.

fishjam:
And I found my previous use the `\r\n` as line ending, it is different from your code. So I change to `\n`.

bgianfo:
Thanks for the patch!

Issue #240, opened 2022-06-07
build error: 'GetThreadLocale': undeclared identifier


michaellongge163:
I downloaded newest release Detours-4.01,and run Developer Command Prompt for VS 2022
Then nmake get error
```
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.2.3
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************

D:\VS2022>cd /d D:\GitHub\vmthook\Detours-4.0.1

D:\GitHub\vmthook\Detours-4.0.1>nmake

Microsoft (R) Program Maintenance Utility Version 14.32.31329.0
Copyright (C) Microsoft Corporation.  All rights reserved.

        cd "D:\GitHub\vmthook\Detours-4.0.1"
        cd "D:\GitHub\vmthook\Detours-4.0.1\src"
 Created ..\include
 Created ..\lib.X86
 Created ..\bin.X86
 Created obj.X86
        cl /W4 /WX /Zi /MT /Gy /Gm- /Zl /Od /DWIN32_LEAN_AND_MEAN /D_WIN32_WINNT=0x501 /Fd..\lib.X86\detours.pdb /Foobj.X86\ /c detours.cpp modules.cpp disasm.cpp image.cpp creatwth.cpp disolx86.cpp disolx64.cpp disolia64.cpp disolarm.cpp disolarm64.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.32.31329 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

detours.cpp
modules.cpp
disasm.cpp
image.cpp
creatwth.cpp
disolx86.cpp
disolx64.cpp
disolia64.cpp
disolarm.cpp
disolarm64.cpp
Generating Code...
        link /lib /out:..\lib.X86\detours.lib obj.X86\detours.obj      obj.X86\modules.obj      obj.X86\disasm.obj       obj.X86\image.obj        obj.X86\creatwth.obj     obj.X86\disolx86.obj     obj.X86\disolx64.obj     obj.X86\disolia64.obj    obj.X86\disolarm.obj     obj.X86\disolarm64.obj
Microsoft (R) Library Manager Version 14.32.31329.0
Copyright (C) Microsoft Corporation.  All rights reserved.

        copy detours.h ..\include\detours.h
        1 file(s) copied.
        copy detver.h ..\include\detver.h
        1 file(s) copied.
[...]

Issue #202, opened 2021-05-14
building issue GetThreadLocale': undeclared identifier


djn3m0:
Hi,

Im trying to build Microsoft Detours for x64 and this is the error I get, I managed to build this correctly few days ago but now I cannot build it again! I couldn't find what could be wrong :(

```
SET DETOURS_TARGET_PROCESSOR=X64
"C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat"
NMAKE

Microsoft (R) Program Maintenance Utility Version 14.28.29915.0
Copyright (C) Microsoft Corporation.  All rights reserved.

        cd "C:\Users\User1\Desktop\Detours-4.0.1\Detours-4.0.1"
        cd "C:\Users\User1\Desktop\Detours-4.0.1\Detours-4.0.1\src"
 Created ..\include
 Created ..\lib.X64
 Created ..\bin.X64
 Created obj.X64
        cl /W4 /WX /Zi /MT /Gy /Gm- /Zl /Od /DWIN32_LEAN_AND_MEAN /D_WIN32_WINNT=0x501 /Fd..\lib.X64\detours.pdb /Foobj.X64\ /c detours.cpp modules.cpp disasm.cpp image.cpp creatwth.cpp disolx86.cpp disolx64.cpp disolia64.cpp disolarm.cpp disolarm64.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.28.29915 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.
...
...
...
Microsoft (R) Windows (R) Resource Compiler Version 10.0.10011.16384
Copyright (C) Microsoft Corporation.  All rights reserved.

        cl /nologo /Zi /MT /Gm- /W4 /WX /Od /I..\..\include /Fdobj.X64\vc.pdb /Foobj.X64\ /c trcapi.cpp
trcapi.cpp
C:\Users\User1\Desktop\Detours-4.0.1\Detours-4.0.1\samples\traceapi\_win32.cpp(4323): error C2065: 'GetThreadLocale': undeclared identifier
C:\Users\User1\Desktop\Detours-4.0.1\Detours-4.0.1\samples\traceapi\_win32.cpp(7050): error C2065: 'SetThreadLocale': undeclared identifier
[...]

PR #247, opened 2022-07-23
fix build error for error C2065: 'GetThreadLocale'

fishjam:
When I try to build sample in vcpkg's download detours code(v4.0.1-2f7d798f9a) , I meet same `build error: C2065: 'GetThreadLocale'` as #202  and #240 , although there is [fix](https://github.com/microsoft/Detours/commit/c89449a12470e7efe0518d8af062c934f636f1b9#diff-446d52d5e28f864320cf9ebb7acdc538566de36a09cab28050c834ea0eb09273R4322) by add `#if(WINVER >= 0x0500)`, but I think it's not best solution now, especially for traceapi( it can not hook GetThreadLocale ). 

consider detour src build configuration is [`D_WIN32_WINNT=0x501`](https://github.com/microsoft/Detours/blob/master/src/Makefile#L24) now, so update the samples's build macro to high value, 0x0501(same as source code define macro)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/247)

### DIFF
--- a/samples/traceapi/trcapi.cpp
+++ b/samples/traceapi/trcapi.cpp
@@ -7,7 +7,7 @@
 //  Copyright (c) Microsoft Corporation.  All rights reserved.
 //
 #undef WIN32_LEAN_AND_MEAN
-#define _WIN32_WINNT        0x400
+#define _WIN32_WINNT        0x0501
 #define WIN32
 #define NT
 #define _WINSOCK_DEPRECATED_NO_WARNINGS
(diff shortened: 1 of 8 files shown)


sylveon:
The source code builds with Windows XP as target, and does use APIs introduced in Windows XP, so the idea of Windows NT compat is probably gone already. This just fixes up the samples to also target Windows XP.

bgianfo:
> The source code builds with Windows XP as target, and does use APIs introduced in Windows XP, so the idea of Windows NT compat is probably gone already. This just fixes up the samples to also target Windows XP.

That's true, I had misremembered the `_WIN32_WINNT` version `src\Makefile` specifies.
Thanks for keeping me honest!

bgianfo (approved):

PR #257, opened 2022-09-26
Add context callback support for x86 and x64

JohnMcPMS:
Adds a context callback function that is invoked before the detour itself.

Only supports x86 and x64, as that is all I had in the source of the port.
Is this acceptable?  It should fail when attempted on another architecture in this PR; does it need to work on all supported architectures?

The additional bytes required for every trampoline are likely to be viewed as a problem.  Because of that, I made it a conditional compilation feature.  The interface remains the same, but if one attempts to pass in a callback without having compiled in support, the detour attach fails.

Other issues that @jaykrell pointed out, but don't seem to be necessary for the shipping use case:
1. The "trampoline" only preserves the integer fastcall registers, not the floating point registers.  Since the shipping use case is to call `TlsSetValue`, this doesn't end up being an issue.
2. No pdata/xdata equivalents are created (presumably via `RtlAddFunctionTable`) for the "trampoline".  This doesn't appear to be an issue for the shipping use case, but I agree that it is undesirable for general use.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/257)


.github/workflows/main.yml
```diff
@@ -3,6 +3,8 @@ name: CI-Build
 env:
   # Turn on msvc analysis during build, enable once warnings are clean.
   DETOURS_ANALYZE: true
+  # Enable feature for CI build
+  DETOURS_SUPPORT_CONTEXT_CALLBACK: true
```
sylveon, line 7: This should probably be a test matrix, to keep the non callback path tested.

JohnMcPMS: I was hesitant to double the number of workflows that were run (from 16 to 32).

I changed it to run the matrix, but exclude both ARM flavors as they are not supported (so 24 workflows).  The exclude can be removed easily if that is desired.

src/detours.cpp
```diff
@@ -122,14 +122,42 @@ struct _DETOUR_TRAMPOLINE
...
+C_ASSERT(sizeof(_DETOUR_TRAMPOLINE) == 96);
+#else
 C_ASSERT(sizeof(_DETOUR_TRAMPOLINE) == 72);
+#endif // DETOURS_SUPPORT_CONTEXT_CALLBACK
 
 enum {
     SIZE_OF_JMP = 5
 };
 
+#ifdef DETOURS_SUPPORT_CONTEXT_CALLBACK
```
jaykrell, line 141: "SUPPORT" => "ENABLE" maybe? I don't know.

src/detours.cpp
```diff
@@ -334,14 +362,72 @@ struct _DETOUR_TRAMPOLINE
...
+    *pbCode++ = 0x51;
+
+    *pbCode++ = 0x48;   // sub rsp,20h (required by x64 __fastcall)
+    *pbCode++ = 0x83;
+    *pbCode++ = 0xEC;
+    *pbCode++ = 0x20;
+
+    *pbCode++ = 0x48;   // mov rcx,imm64 (param)
+    *pbCode++ = 0xB9;
+    *((INT64*&)pbCode)++ = (INT64)(pContextParam);
```
jaykrell, line 408: nit: No need for the last set of parens.

src/detours.cpp
```diff
@@ -334,14 +362,72 @@ struct _DETOUR_TRAMPOLINE
...
+    *pbCode++ = 0x83;
+    *pbCode++ = 0xC4;
+    *pbCode++ = 0x20;
+
+    *pbCode++ = 0x41;   // pop r9 (restore r9)
+    *pbCode++ = 0x59;
+    *pbCode++ = 0x41;   // pop r8 (restore r8)
+    *pbCode++ = 0x58;
+    *pbCode++ = 0x5A;   // pop rdx (restore rdx)
+    *pbCode++ = 0x59;   // pop rcx (restore rcx)
```
jaykrell, line 424: Likely an ABI violation.
Detours already violates ABI.
This likely makes it worse.
I understand it is under ifdef and that makes it far more acceptable, but I do wonder how it impacts your scenario too.

Does debugging successfully show stack for you?
Have you any scenarios that raise an exception with this code on stack?

Would it be better to have kinda templatized C detours to fetch the context from thread locals?

JohnMcPMS: For our use, the context function is always this:
```C++
void __stdcall set_hook_context(void* context)
{
	if (s_hook_context_tls_index_ != TLS_OUT_OF_INDEXES)
		TlsSetValue(s_hook_context_tls_index_, context);
}
```
`TlsSetValue` could theoretically raise an exception, although I don't think the current implementation would for an allocated index.

I don't know about debugging; this is a legacy component for us and it would take some time to figure that out.

As I understand it (which is admittedly not thoroughly), the goal is to effectively add another parameter to the detoured function.  The value of this parameter is determined at attach time, not compile time.  I have not been able to think of another way to handle the scenario that doesn't involve emitting code at runtime.  I suppose with enough effort (and possible assistance from the attach caller), one could actually implement this as adding another parameter rather than stashing it in TLS.  I'm not sure how difficult/feasible that would be, nor can I predict the potential impact.

Is the primary concern the lack of function table entries?  Is there enough referential data to implement the callback for `RtlInstallFunctionTableCallback`, or would that require significant investment?

jaykrell: Ideally you could get stackwalks in a debugger too, but you appear to break that. I guess nobody really cares.

src/detours.cpp
```diff
@@ -334,14 +362,72 @@ struct _DETOUR_TRAMPOLINE
...
+
+inline PBYTE detour_gen_context_callback(PBYTE pbCode, PF_DETOUR_INVOKE_CONTEXT_CALLBACK* ppfContextCallback, PVOID pContextParam)
+{
+    PBYTE pbJmpSrc = pbCode + 26;
+
+    // Backup volatile registers that are used to pass __fastcall params.
+    // If the set context function performed floating point operations, 
+    // registers XMM0, XMM1, XMM2, and XMM3 would also have to be backed 
+    // up and restored. In the attempt to keep this code generation as simple 
+    // as possible, floating point operations in the set context function isn't
```
jaykrell, line 392: What guarantees the compiler won't use floating point registers??

src/detours.cpp
```diff
@@ -334,14 +362,72 @@ struct _DETOUR_TRAMPOLINE
...
+{
+    PBYTE pbJmpSrc = pbCode + 26;
+
+    // Backup volatile registers that are used to pass __fastcall params.
+    // If the set context function performed floating point operations, 
+    // registers XMM0, XMM1, XMM2, and XMM3 would also have to be backed 
+    // up and restored. In the attempt to keep this code generation as simple 
+    // as possible, floating point operations in the set context function isn't 
+    // currently supported.
+    *pbCode++ = 0x51;   // push rcx (backup rcx)
```
jaykrell, line 394: Is this the start of a function?
Are you aware of the ABI detail that functions must start with instructions that are at least one byte?
Well, it is for hotpatch, maybe no matter.

src/detours.cpp
jaykrell: You are aware of TlsGetValue / TlsSetValue's tendency to clear last error, I assume?

jaykrell: You should try to use declspec(thread).

PR #261, opened 2022-10-21
Enable Detours to be used without allocating or freeing memory while threads are suspended.

PDeets:
Detours uses `new` and `delete` while threads are suspended. This can cause deadlocks if the suspended thread has locked a mutex in the heap. (See issue #70.) Additionally, when AppVerifier is enabled, `VirtualFree` and `VirtualAlloc` can be hooked by AppVerifier. These hooks do memory operations on the heap internally. The deadlocks are easier to hit when AppVerifier is enabled.

Back in 2014, I worked with https://github.com/KnicKnic on a fix for this. We also consulted with https://github.com/galenh on the API design changes. I checked in the fix to a copy of Detours in the Windows repository within Microsoft, but this wasn't propagated back to the Detours master sources. I'm now bringing a cleaned-up version of the fix through GitHub so my code in the Windows repo can switch to the open-source version of Detours.

The fix makes it so all calls to `free` and `VirtualFree` are done after all calls to `ResumeThread` in `DetourTransactionAbort` and `DetourTransactionCommit`. However, to fully avoid deadlocks, the user needs to also do the following:
* Switch from `DetourUpdateThread` to the new `DetourUpdateThreadPreallocated` API.
* Make all calls to `DetourAttach`, `DetourAttachEx`, and `DetourDetach` before calling `DetourUpdateThreadPreallocated` in a given transaction.
* Avoid allocating or freeing any memory in user code between the first call to `DetourUpdateThreadPreallocated` and the call to `DetourTransactionAbort` or `DetourTransactionCommit`.

If you follow those rules, you will avoid memory allocations while threads are suspended in the process.

The `DetourUpdateThreadPreallocated` API is like `DetourUpdateThread` except the user passes in a `DETOUR_THREAD_DATA` struct that they allocated themselves ahead of time. The user is responsible for freeing the memory after aborting or committing the transaction. In my code, I store the `DETOUR_THREAD_DATA` instances in a `std::vector` that I resize ahead of time in order to front-load the allocation before starting the detour transaction.

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -1506,8 +1506,10 @@ static BOOL detour_is_region_empty(PDETOUR_REGION pRegion)
     return TRUE;
 }
 
-static void detour_free_unused_trampoline_regions()
+static PDETOUR_REGION detour_splice_unused_trampoline_regions()
 {
+    PDETOUR_REGION pRegionsToFree = NULL;
+
     PDETOUR_REGION *ppRegionBase = &s_pRegions;
     PDETOUR_REGION pRegion = s_pRegions;
 
@@ -1515,7 +1517,10 @@ static void detour_free_unused_trampoline_regions()
         if (detour_is_region_empty(pRegion)) {
             *ppRegionBase = pRegion->pNext;
 
-            VirtualFree(pRegion, 0, MEM_RELEASE);
+            // Move pRegion to the pRegionsToFree list.
+            pRegion->pNext = pRegionsToFree;
+            pRegionsToFree = pRegion;
+
             s_pRegion = NULL;
         }
         else {
@@ -1523,16 +1528,12 @@ static void detour_free_unused_trampoline_regions()
         }
         pRegion = *ppRegionBase;
     }
+
+    return pRegionsToFree;
 }
 
 ///////////////////////////////////////////////////////// Transaction Structs.
 //
-struct DetourThread
-{
-    DetourThread *      pNext;
-    HANDLE              hThread;
-};
-
 struct DetourOperation
 {
     DetourOperation *   pNext;
@@ -1549,7 +1550,8 @@ static BOOL                 s_fRetainRegions        = FALSE;
 static LONG                 s_nPendingThreadId      = 0; // Thread owning pending transaction.
 static LONG                 s_nPendingError         = NO_ERROR;
 static PVOID *              s_ppPendingError        = NULL;
-static DetourThread *       s_pPendingThreads       = NULL;
+static DETOUR_THREAD_DATA * s_pPendingThreads       = NULL;
+static DETOUR_THREAD_DATA * s_pUnownedPendingThreads = NULL; // Pending threads where the user allocated the DETOUR_THREAD_DATA instead of us.
 static DetourOperation *    s_pPendingOperations    = NULL;
 
 //////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


src/detours.cpp
```diff
@@ -1838,76 +1919,16 @@ LONG WINAPI DetourTransactionCommitEx(_Out_opt_ PVOID **pppFailedPointer)
...
+        detour_move_thread_instruction_pointer_if_needed(t->hThread);
     }
 
     // Restore all of the page permissions and flush the icache.
     HANDLE hProcess = GetCurrentProcess();
-    for (o = s_pPendingOperations; o != NULL;) {
+    for (o = s_pPendingOperations; o != NULL; o = o->pNext) {
         // We don't care if this fails, because the code is still accessible.
         DWORD dwOld;
         VirtualProtect(o->pbTarget, o->pTrampoline->cbRestore, o->dwPerm, &dwOld);
```
PDeets, line 1934: Note that AppVerifier can also hook `VirtualProtect`, but I carefully reviewed the code for its hook and it never acquires any locks. I think back in 2014 when this change was originally made, the AppVerifier hook for `VirtualProtect` did acquire locks, so we moved the `VirtualProtect` calls to be after the `ResumeThread` calls. However, I didn't include that here since I don't think it is needed with the current version of AppVerifier's `VirtualProtect` hook. Also, for safety reasons, it is good to remove writability from executable pages before we resume threads.

The code for `VirtualProtect` itself can acquire a lock if we are restoring to read_execute and something has called `MmSecureVirtualMemory` with `PAGE_READWRITE`. However, this is very unlikely to happen in the case of Detours. As I understand it, `MmSecureVirtualMemory` with `PAGE_READWRITE` can be used by a driver if you pass it a user-mode buffer that it needs to fill in from kernel-mode. The API would ensure user-mode code can't switch the page to be read-only before the kernel-mode code gets a chance to write to the buffer. Executable code should not be used as the destination for a write from a driver, so I'm OK with a deadlock if this happens.

jaykrell:
Aside: For VirtualAlloc to turn around and do a heap allocation, is imho bogus. It breaks the layering of the system.

PDeets:
> Aside: For VirtualAlloc to turn around and do a heap allocation, is imho bogus. It breaks the layering of the system.

I get that, but it is only when you opt-in by enabling AppVerifier. AppVerifier does allocations to track usage. This supports the !avrf debugger extension which can show the history of callstacks for calls to VirtualAlloc and VirtualFree. This history can be very useful for tracking down the root cause of a bug when AppVerifier catches one.

PDeets:
> > Aside: For VirtualAlloc to turn around and do a heap allocation, is imho bogus. It breaks the layering of the system.
> 
> I get that, but it is only when you opt-in by enabling AppVerifier. AppVerifier does allocations to track usage. This supports the !avrf debugger extension which can show the history of callstacks for calls to VirtualAlloc and VirtualFree. This history can be very useful for tracking down the root cause of a bug when AppVerifier catches one.

Actually, I looked over the code to fact-check this. I think I'm wrong. The AppVerifier hooks appears to avoid allocations on the process heap. It uses a different set of lower-level allocation APIs. Regardless though, the implementation acquires locks in various places. We still need to avoid VirtualAlloc and VirtualFree calls with suspended threads while AppVerifier is enabled in order to avoid deadlocks due to the various locks that are acquired in these hooks.

pkaku (approved):

PR #277, opened 2023-03-24
Added libdetours that generates detours.dll

lostmsu:
Exported functions, that inject DLLs into other processes (for eventual consumption from .NET)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/277)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/277)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/Detours/pull/277)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,8 @@ all:
     @if exist "$(MAKEDIR)\core\makefile" cd "$(MAKEDIR)\core" && $(MAKE) /NOLOGO /$(MAKEFLAGS)
     cd "$(MAKEDIR)\src"
     @$(MAKE) /NOLOGO /$(MAKEFLAGS)
+    cd "$(MAKEDIR)\libdetours"
+    @$(MAKE) /NOLOGO /$(MAKEFLAGS)
     cd "$(MAKEDIR)\samples"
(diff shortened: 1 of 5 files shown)


sylveon:
I believe making detours a static library only is an intentional choice. Why not simply expose ways to call these Detours APIs directly in the DLL you're trying to inject?

lostmsu:
@sylveon that would bring all the DLL's dependencies into the process that wants to perform injection.

sylveon:
Injected DLLs should already strive to as little deps as possible.

lostmsu:
@sylveon I mean yes, that's the ideal, but having the host to be compatible with all target related dependencies is not always achievable.

The other scenario would be the need to inject multiple DLLs or inject DLLs based on runtime information (mine is close to this - I am trying to detour APIs called by games, and the set of hooks varies by game).

PR #307, opened 2023-11-06
update system DLL range logic for 64-bit OSes

jdu2600:
Hey Detours folks,

Just a couple of suggestions for choosing the trampoline location.

The X86 range reserved for system DLLs is not up to date.

The X64 "not +/- 1GB of target" heuristic to avoid the range used by system DLLs is not entirely robust.  Instead we could replace this with a "not 1GB after ntdll" heuristic.

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -53,8 +53,49 @@ C_ASSERT(sizeof(_DETOUR_ALIGN) == 1);
 //
 // Region reserved for system DLLs, which cannot be used for trampolines.
 //
-static PVOID    s_pSystemRegionLowerBound   = (PVOID)(ULONG_PTR)0x70000000;
-static PVOID    s_pSystemRegionUpperBound   = (PVOID)(ULONG_PTR)0x80000000;
+// On Windows 10, ntdll.dll is mapped at random location (ASLR) within a
+// range near the top of the usermode address space.
+// After that, further system DLLs are mapped top down within this range.
+// In the bottom of the range is reached, the allocator wraps to the top.
+//
+// We also want to exlcude any pages that share a CFG bitmap page with a system DLL
+// so leave an additional a 1MB buffer on each side of the range.
+//
+#if defined(DETOURS_64BIT)
+// On X64 the range is 0x7FF800000000..0x7FFFFFFF0000 - which is 32GB!
+// So we likely must allocate in the system DLL range to be +/- 2GB.
+// But we want to avoid at least the first 1GB that will be used for system DLLs.
+// Due to wrapping, this may be two seperate ranges.
+static PVOID    s_pSystemRegionUpperBound = (PVOID)((ULONG_PTR)GetModuleHandleW(L"ntdll.dll") + (ULONG_PTR)0x100000);
+static PVOID    s_pSystemRegionLowerBound = (PVOID)((ULONG_PTR)s_pSystemRegionUpperBound < (ULONG_PTR)0x7FF83F000000 ?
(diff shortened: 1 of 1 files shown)


Mauler125:
This breaks support for Windows 7, `DetourAllocateRegionWithinJumpBounds` emits the message "Couldn't find available memory region!\n".

jdu2600:
@Mauler125 Was this Windows 7 x86?
The x86 logic was updated to account for Windows 8 kernel changes - but I didn't consider backwards compatibility with EOL versions at time.

https://github.com/microsoft/Detours/blob/4ed81f1036d028306b3659e976d2e57f37c4245a/src/detours.cpp#L78-L84

Mauler125:
@jdu2600 This was on the latest version of Windows 7 x64 (SP1 with all updates), I think the ranges are different on Windows 7, and changed in Windows 8 as you mentioned. I hadn't had the chance to debug this myself yet. Some of our users host a game server on Windows 7 x64 which is where this was being reported from.

These changes however do seem to work perfectly on Windows 10 and Windows 11 for the 2 months we have it implemented and tested; not a single issue has been encountered from it yet.

RatinCN:
> @jdu2600 This was on the latest version of Windows 7 x64 (SP1 with all updates), I think the ranges are different on Windows 7, and changed in Windows 8 as you mentioned. I hadn't had the chance to debug this myself yet. Some of our users host a game server on Windows 7 x64 which is where this was being reported from.
> 
> These changes however do seem to work perfectly on Windows 10 and Windows 11 for the 2 months we have it implemented and tested; not a single issue has been encountered from it yet.

Weird, I've checked "Windows Internals 6th" (corresponding to Win7) and analyzed `ntoskrnl.exe!MiInitializeRelocations`, the results show the region are the same SINCE Win7:
32-bit process：[0x50000000 ... 0x78000000], a total of 640MB
64-bit process：[0x00007FF7FFFF0000 ... 0x00007FFFFFFF0000], a total of 32GB

RatinCN:
In Win7, ASLR can be turned off by registry:
> HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Memory Management, MoveImages (REG_DWORD)

I think there's a chance this could be causing this issue.

Mauler125:
> In Win7, ASLR can be turned off by registry:
> 
> > HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Memory Management, MoveImages (REG_DWORD)
> 
> I think there's a chance this could be causing this issue.

Sorry for my later responce. Today I did some research again on this issue and it was indeed caused by having ASLR disabled system-wide using the mentioned registry value.

Your recent changes on KNSoft's SlimDetours does fix the issue as far as I've been able to test it. I will continue testing it and report any issues if I encounter them.

RatinCN:
> > In Win7, ASLR can be turned off by registry:
> > > HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Memory Management, MoveImages (REG_DWORD)
> > 
> > 
> > I think there's a chance this could be causing this issue.
> 
> Sorry for my later responce. Today I did some research again on this issue and it was indeed caused by having ASLR disabled system-wide using the mentioned registry value.
> 
> Your recent changes on KNSoft's SlimDetours does fix the issue as far as I've been able to test it. I will continue testing it and report any issues if I encounter them.

Thanks, if still encounter the same problem, try to remove [this block](https://github.com/KNSoft/KNSoft.SlimDetours/blob/a06830e01c4d1c381d88a74350ba1ec8860a6bd7/Source/SlimDetours/Memory.c#L134-L144), I don't have 100% confidence in this.

PR #308, opened 2023-11-06
don't break stack walking

jdu2600:
Detouring a function should not break our ability to walk the call stack.

On X64, suggest that we require the Detour to reside in MEM_IMAGE so that Windows has access to the compiler generated function information necessary for unwinding.  This will prevent surprises.

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -2023,6 +2023,27 @@ LONG WINAPI DetourAttachEx(_Inout_ PVOID *ppPointer,
         return ERROR_INVALID_PARAMETER;
     }
 
+#ifdef DETOURS_X64
+    // Detouring a function should not break our ability to walk the call stack.
+    // On X64 stack unwinding relies on RUNTIME_FUNCTION structures.
+    // These are typically found in the .pdata section of PE32+ images.
+    // Require this.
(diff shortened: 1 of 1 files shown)


src/detours.cpp
```diff
@@ -2023,6 +2023,22 @@ LONG WINAPI DetourAttachEx(_Inout_ PVOID *ppPointer,
...
+#ifdef DETOURS_X64
+    // Detouring a function should not break our ability to walk the call stack.
+    // On X64 stack unwinding relies on RUNTIME_FUNCTION structures.
+    // Require this.
+    // These are typically found in the .pdata section of PE32+ images.
+    // Alternatively dynamic code should use RtlAddGrowableFunctionTable
+    // to register this information. This supports kernel stack walks.
+    // See https://learn.microsoft.com/en-us/cpp/build/exception-handling-x64
+    DWORD64 imageBase;
+    if (NULL == RtlLookupFunctionEntry((DWORD64)pDetour, &imageBase, NULL)) {
```
kobykahane, line 2035: In some cases, A Detour function could legitimately be a leaf function and thus have no function table entry, so having this requirement is somewhat over-strict.
Regardless, this shouldn't be just `DETOURS_X64` specific - function table unwinding is also used on ARM64, for example.

PR #311, opened 2024-01-10
Update main.yml to disable extraneous permissions.

galenh:


BissetJ:
I added explicit permissions to get CodeQL working again, that obsoletes this PR.

PR #313, opened 2024-03-28
Work-around an issue in Arm64 regarding the isolated use of CONTEXT_CONTROL.

pmsjt:
Work-around an issue in Arm64 (and Arm64EC) in which LR and FP registers may become zeroed when CONTEXT_CONTROL is used without CONTEXT_INTEGER.

The addition of the CONTEXT_INTEGER flag does not translate to any real overhead: The kernel performs relatively expensive stack unwind operations for Get and Set ThreadContext touching and scanning over large unwind info datastructures. In scenario the extra few integers register copies (which share cache lines with CONTEXT_CONTROL) can't be observed in the scale and the noise.

This change is adding CONTEXT_INTEGER to the Get and SetThread context calls for both Arm64 and x86_64 so that Arm64EC is also addressed.

This issue is being addressed in the OS as well, but it will take time to disseminate the fix and backport it to all released versions. When the fix is out, this change can arguably be considered redundant, but given there is no real downside, I don't see an objective reason to not add it or making diligent plans to remove it.

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -1837,35 +1837,40 @@ LONG WINAPI DetourTransactionCommitEx(_Out_opt_ PVOID **pppFailedPointer)
         }
     }
 
-    // Update any suspended threads.
-    for (t = s_pPendingThreads; t != NULL; t = t->pNext) {
-        CONTEXT cxt;
-        cxt.ContextFlags = CONTEXT_CONTROL;
-
 #undef DETOURS_EIP
+#undef DETOURS_CONTEXT_FLAGS
 
 #ifdef DETOURS_X86
 #define DETOURS_EIP         Eip
+#define DETOURS_CONTEXT_FLAGS CONTEXT_CONTROL
 #endif // DETOURS_X86
 
 #ifdef DETOURS_X64
 #define DETOURS_EIP         Rip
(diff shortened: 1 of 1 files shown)


src/detours.cpp
```diff
@@ -1837,35 +1837,40 @@ LONG WINAPI DetourTransactionCommitEx(_Out_opt_ PVOID **pppFailedPointer)
...
+#undef DETOURS_CONTEXT_FLAGS
 
 #ifdef DETOURS_X86
 #define DETOURS_EIP         Eip
+#define DETOURS_CONTEXT_FLAGS CONTEXT_CONTROL
 #endif // DETOURS_X86
 
 #ifdef DETOURS_X64
 #define DETOURS_EIP         Rip
+#define DETOURS_CONTEXT_FLAGS (CONTEXT_CONTROL | CONTEXT_INTEGER)
```
jaykrell, line 1850: Should we "just" use this consistently for all architecture?
i.e. also x86?

jaykrell (approved):

src/detours.cpp
jaykrell: Add comment explaining it? I realize there is git blame and PRs. Some orgs are super comment heavy these days and want everything explained on every line. I realize Detours is not that org.

BissetJ (approved):

PR #318, opened 2024-07-22
Fix detours ability to identify OS applied patches.

rishaw96:
This change allows for interoperability between OS applied patches and detours (x64/arm64/x86). When a detour is active and an OS patch is attempted to be applied, the attempt will fail since it's not safe/secure to patch code amidst unexpected code stream. However, it is safe for the detours library to insert a detour into a function that has already been patched by the OS. In the event a binary/process is already detoured and a patch needs to be applied, the process can simply be restarted to get the patch applied prior to the detour initialization. 

Amd64:
Prior to this change, when a patched binary has a detour applied to it, the detour application will incorrectly redirect to the detour at the point of the HPAT code page jump. This results in failure to insert the detour code, since the detours library cannot modify the HPAT region. Now, the detours library first verifies that it has encountered an OS applied patch function, and if so, it redirects execution to the detour at the point of the 6-byte function padding leading up to the patched function. 

Arm64:
Similar to the amd64 issue, the detour code was interpreting the patch function jump as an end-of-function instruction, resulting in failure to apply the detour. The detour library now verifies that the code encountered is an OS patch and modifies the bytes at the start of the patched function.  

x86:
Same as amd64, with a slight deviation in expected code stream for OS applied patches. 

One other minor change was needed in uimports.cpp where the NT header checksum field was being zeroed, leading to patch applicability failing (since the OS uses file checksum/timestamp fields to identify a patch).

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -156,6 +156,8 @@ inline PBYTE detour_gen_brk(PBYTE pbCode, PBYTE pbLimit)
 
 inline PBYTE detour_skip_jmp(PBYTE pbCode, PVOID *ppGlobals)
 {
+    PBYTE pbCodeOriginal;
+
     if (pbCode == NULL) {
         return NULL;
     }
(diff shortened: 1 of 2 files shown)


galenh (approved):
Nice.

PR #325, opened 2024-11-02
Fixed detour_alloc_trampoline_allocate_new

xorcus:
detour_alloc_trampoline_allocate_new(pbTarget, pLo, pHi) was designed with the assumption that pbTarget is going to be between pLo and pHi. In our case, 32-bit cscript!mainCRTStartup was loaded at 0x34020, i.e. pbTarget was below pLo (which was fixed to 0x80000 in detour_2gb_below), and therefore:
- detour_alloc_region_from_hi(pLo, pbTarget) did not do anything
- detour_alloc_region_from_lo(pbTarget, pHi) allocated a 64 KiB block below pLo (in our case happened to be 0x70000) which was later discarded and NULL returned in detour_alloc_trampoline 

The fix clamps pbTarget into [pLo, pHi] range.

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -1410,6 +1410,21 @@ static PVOID detour_alloc_region_from_hi(PBYTE pbLo, PBYTE pbHi)
     return NULL;
 }
 
+// Return pbTarget clamped into [pLo, pHi] range.
+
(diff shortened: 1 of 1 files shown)


xorcus:
@microsoft-github-policy-service agree company="Microsoft"

PR #341, opened 2025-04-01
CI Fixups

BissetJ:
* codeql-action and upload-artifact need to be bumped to newer versions.
* The artifact names must be unique across jobs, upload-artifact v4 won't merge all the outputs into a single archive like before.


BissetJ:
The 32-bit ARM builds are still failing on windows-2022, but this seems to fix the other CI issues.

PR #342, opened 2025-04-02
CI: Disable 32-bit ARM in the CI builds.

BissetJ:
The 32-bit ARM builds no longer succeed with the latest platform SDK because 32-bit ARM support has been deprecated and there are no longer libs for things like ws_32.  This is why the windows-2022 CI builds are failing for the two 32-bit ARM jobs.  32-bit ARM isn't much of a concern for anybody, I'm just going ahead and disabling it for the CI runs for now.


BissetJ:
Note that this is not removing 32-bit ARM support from Detours, just (temporarily) from the CI builds.

PR #352, opened 2025-06-20
Fix some bugs in samples

gotoh:
Fix two tiny bugs in sample `tracetcp`.
- Unexpected behavior by unnecessary `__debugbreak()` in `Mine_WSALookupServiceBeginW()`.
- Crash in `Mine_accept()` when the *optional* second argument `addr` of `accept()` is `NULL`.
- Fix misjudgement with return value from `GetTimeZoneInformation()`.


gotoh:
@microsoft-github-policy-service agree

BissetJ (approved):

